PR #52, opened 2016-05-20
added some additional input_cmd_funcs and named tuple Coordinate.

obtitus:
This is not complete, there are still missing commands and not all of the functions return a Coordinate object. But I just wanted to know if this looks like a sensible approach? 

I attached a cylinder_2D_py.in, which is definitely more verbose than the existing example, but I prefer the explicitness of the python syntax.

Something should also be done to link the documentation to the input.html, I have just copy pasted most of the doc-strings from there...


craig-warren:
Thank @obtitus that will definitely be a help to those building input files with Python!

PR #55, opened 2016-06-02
Array job mode

jnally-dstl:
Added to provide an alternative method for running in parallel on an HPC cluster through a job scheduler which supports it. This includes additional flags:
"-array" to specify that this is an array job
"--array-id" to specify the current task id

The "-n" flag is also required to define the total number of model runs.

I note that Taguchi optimisation appears not to work but testing with other modes suggests yielded the same results so I did not remove the option.

### DIFF
--- a/gprMax/gprMax.py
+++ b/gprMax/gprMax.py
@@ -1,6 +1,8 @@
 # Copyright (C) 2015-2016: The University of Edinburgh
 #                 Authors: Craig Warren and Antonis Giannopoulos
 #
+# The contents include material subject to (C) Crown copyright (2016), Dstl.
+#
 # This file is part of gprMax.
 #
 # gprMax is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


craig-warren:
@jnally-dstl Great, thanks. I think people will find this useful. 

Before I merge the changes, I wonder if it even simpler to just have the ``--array-id`` command line argument, and then only modify the ``run_std_sim`` function. Inside the ``run_std_sim`` function ``modelrun`` would be set by ``--array-id`` (if present) and otherwise automatically from the for loop. What are your thoughts?

Also, can you provide a snippet from the job submission script to show how you are running gprMax as an array job.

jnally-dstl:
@craig-warren that is good to hear.

I did contemplate using the `run_std_sim` function but I decided that it might be best to define the array task as an easily identifiable mode to improve readability. I would also have had to duplicate the code within the loop anyway unless I split it out into a separate function.

craig-warren:
@jnally-dstl can you email me (Craig.Warren at ed.ac.uk) so we can have a chat about the merge. Thanks.

PR #91, opened 2017-01-25
bugfix, to_save_str = ' '.join(to_save)  fails if to_save is Noen

obtitus:


obtitus:
ehh, sorry about adding multiple commits to this request, I hate git... The first commit is a bug-fix, while the next 2 are a start on some unit-testing of the input_cmd_funcs file.

craig-warren:
@obtitus thanks thats great. I will move the test into the test sub-package. I've done a bit of work in there for testing at a model level, but more unit testing is *definitely* welcome!

PR #103, opened 2017-04-12
Added new input file command to specify interpolated user input waveforms

rsawtell:
Use of the #excitation_file command is tricky because it does not specify the time domain in conjunction with waveform amplitude values. 

Even when using the exact same functions as standard gprMax waveforms, as well as a timestep value pulled from an output file with the same domain configuration, the resulting output file will be excessively unstable as seen below:

usergauss

Compared to the standard gaussian waveform of 2Ghz:
usergauss3

I wasn't sure the best approach to modifying this, so I put my changes in a new input command, #excitation_filefunc, which uses scipy UnivariateSplines to interpolate input data allowing the simulations to query the waveform at arbitrary time. Using the exact same input file as the first graph, but with the time domain specified and using this interpolation method results in vastly improved results:
usergauss2

Hopefully you find this helpful. I've attached the input files and excitation files used below so you can replicate it yourself:
[usertest.zip](https://github.com/gprMax/gprMax/files/917697/usertest.zip)

### DIFF
--- a/gprMax/waveforms.py
+++ b/gprMax/waveforms.py
@@ -24,7 +24,7 @@ from gprMax.utilities import round_value
 class Waveform(object):
     """Definitions of waveform shapes that can be used with sources."""
 
-    types = ['gaussian', 'gaussiandot', 'gaussiandotnorm', 'gaussiandotdot', 'gaussiandotdotnorm', 'ricker', 'sine', 'contsine', 'impulse', 'user']
+    types = ['gaussian', 'gaussiandot', 'gaussiandotnorm', 'gaussiandotdot', 'gaussiandotdotnorm', 'ricker', 'sine', 'contsine', 'impulse', 'user', 'userfunc']
 
     def __init__(self):
         self.ID = None
@@ -32,6 +32,7 @@ class Waveform(object):
         self.amp = 1
         self.freq = None
         self.uservalues = None
+        self.userfunc = None
 
     def calculate_value(self, time, dt):
         """Calculates value of the waveform at a specific time.
@@ -102,6 +103,12 @@ class Waveform(object):
                 ampvalue = 0
             else:
                 ampvalue = self.uservalues[index]
+        elif self.type == 'userfunc':
+                            
+            if self.userfunc is None:
+                ampvalue = 0
+            else:
+                ampvalue = float(self.userfunc(time))
 
         ampvalue *= self.amp
 
--- a/gprMax/input_cmds_file.py
+++ b/gprMax/input_cmds_file.py
@@ -178,7 +178,7 @@ def check_cmd_names(processedlines, checkessential=True):
     essentialcmds = ['#domain', '#dx_dy_dz', '#time_window']
 
     # Commands that there should only be one instance of in a model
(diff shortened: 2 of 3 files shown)


gprMax/waveforms.py
```diff
@@ -102,6 +103,12 @@ def calculate_value(self, time, dt):
                 ampvalue = 0
             else:
                 ampvalue = self.uservalues[index]
+        elif self.type == 'userfunc':
+                            
+            if self.userfunc is None:
+                ampvalue = 0
+            else:
+                ampvalue = float(self.userfunc(time))
```
craig-warren: Does this not rely on an exact match between the absolute time passed into this function, and the time vector given in the excitation file? Which would mean matching two floating point numbers?

rsawtell: userfunc is generated in input_cmds_singleuse.py with a univariate spline, which provides smooth interpolation between all of the data points provided in the input file.

Here is a closeup view of what this looks like in practice:
image

The red dots are the original data points provided by the user input file, while the blue points are from the userfunc sample at a much finer timescale.

If you want to test it yourself:
```
excitationfile = 'waveform_usergauss2.txt'

# Read all waveform values into an array
waveformvalues = np.loadtxt(excitationfile, skiprows=1, dtype=floattype)

timedata = waveformvalues[:, 0]

userfunc = UnivariateSpline(timedata,waveformvalues[:, 1],ext=1,k=2,s=0)

time2 = np.linspace(0,timedata[-1],10000)

plt.plot(time2,userfunc(time2),'b.')
plt.plot(timedata,waveformvalues[:, 1],'ro')
plt.xlim(.45e-9,.55e-9)
plt.ylim(.9,1)
plt.show()
```
Hope this makes sense!

obtitus:
This would be a great addition, I have the same "tricky" problem with excitation_file as well. May i suggest having some optional parameters passed to #excitation_file and wrap around the more general "scipy.interpolate.interp1d" instead of UnivariateSpline?

Suggested documentation draft:

If the input file contains a column labeled "time" (case-insensitive) additional interpolation options can be passed to #excitation_file for interpolation using scipy.interpolate.interp1d. If no time vector is found in the input file the simulations dt will be used.
* ``kind`` Specifies the kind of interpolation as a string (‘none', ‘linear’, ‘nearest’, ‘zero’, ‘slinear’, ‘quadratic’, ‘cubic’ where ‘zero’, ‘slinear’, ‘quadratic’ and ‘cubic’ refer to a spline interpolation of zeroth, first, second or third order) or as an integer specifying the order of the spline interpolator to use. Default is ‘none‘ for no interpolation (time vector is ignored).
* ``fill_value`` (int, float or "extrapolate"), this value will be used to fill in for requested points outside of the data range. If not provided, then the default is 0. If “extrapolate”, then points outside the data range will be extrapolated.

craig-warren:
@obtitus Yes what you propose it along the lines I was thinking. I'll hopefully get some time this week or next to have a further think and implement something.

craig-warren:
@rsawtell @obtitus Not forgotten this, but we have been busy preparing for an updated release in the next few weeks. Should get a chance to look at this again after that is out.

Craig

craig-warren:
Hey guys,

It has been a long time coming, but I finally found some time to look into this. What I have implemented is based on this pull request and your suggestions, so thanks a lot! 

The changes have been made in the latest release, currently v.3.1.2 (https://github.com/gprMax/gprMax/releases/tag/v.3.1.2). Take a look at the revised docs for the `#excitation_file` command - http://docs.gprmax.com/en/latest/input.html#excitation-file

Please give it a test, and let me know your thoughts.

Issue #107, opened 2017-04-25
Would there be interest in a refactoring effort?


fake-name:
Hi There!

I've been looking through the GprMax codebase in hopes of implementing a GPU solver, but I'm having trouble with a lot of the code being kind of "academic" looking.

For example, there are HUGE lines (500 characters!!!!!), lots of  extremely large functions, and just a lot of code that seems like it's just a giant ball of hair.

Would there be any interest in a series of pull requests that try to break the current ball-of-wax structure into more modular objects?

Thanks!

agianno:
Hi Conor,

Thanks for your interest in gprMax. Indeed gprMax is an "academic" code. The GPR research group in Edinburgh that has been working on this for quite sometime - starting with a "spaghetti" code in C written by myself over the last 20 years - is made up by academics, PhD students, post-docs, etc. Although, some of us have been in computational research for a good number of years we are not professional programmers!

Any effort to make the code more robust, modular and easier to maintain and extent is very welcome although sometimes in these types of codes elegance is desirable up to the point that execution speed does not take a massive hit! 

As far as a GPU version is concerned I must say that we do have a CUDA version with very good performance that we have just finished developing for a research project. We are aiming to publish this very soon and most likely in IWAGPR 2017 and at the moment we are in discussions with our sponsors to merge this GPU gprMax branch with the main open source version of gprMax. The intention is there but things have to be done properly and approvals have to be obtained before we do this. So, if you are looking to do a CUDA implementation maybe it is a good idea to wait a little. Alternatively, an OpenCL implementation will be interesting.

Thanks again for your interest to help with gprMax!

Regards

Antonis

fake-name:
Cool. At the moment, I'm literally just doing simple reindentation things, pep-8,  that sort of thing. Ideally, it shouldn't change the execution at all. I'l probably look at reworking the control logic into a series of more modular classes after I've done this (simple whitespace stuff like this is a great way to get a decent overview of the codebase).

I'd say I'm not a professional coder either, but then I realize that I'm somehow getting paid to do horrible things in python, so I guess I am? News to me.

If there's a CUDA implementation, I'll probably just wait. We're interested in doing a few things - CUDA/OpenCL, Variable-size sub-gridding, and being able to distribute the computation of extremely large scenes across multiple computers.

agianno:
[...]

PR #108, opened 2017-04-27
Partial whitespace cleanup.

fake-name:
Mostly changes discussed in https://github.com/gprMax/gprMax/issues/107

Principally, this is a bunch of changes that seek to get the maximum line length under some control. Right now, it's not 100%, mostly because I haven't determined a good way to do inline splitting of some of the very long string output messages, but all the comments at least should be <100 chars-per-line.


fake-name:
Woot!

PR #125, opened 2017-08-30
Multi-Debye Fitting

IraGia:
A python script is added to the users_lib folder with which complex media (Havriliak-Negami, Johnsher, Complex Refractive Index Mixing and arbitrary dielectric data given by user) can be approximated by multi-Debye functions and furthermore implemented to FDTD. 
  An .rst file was also added to the docs folder to provide a documention for the aforemention script.


craig-warren:
@IraGia Thanks, looks great. I will review and merge as soon as possible.

agianno:
@IraGia and @craig-warren I will suggest adding named variables in the function calls as there are too many parameters and without it will become more tricky to use. Not necessary but I think will help.

alberto-battistel:
Hi I do not know if this conversation is still active...
I do not fully understand how to do when I want to model both dielectric permittivity and conductivity. I am dealing with biological tissue where they are both frequency-dependent and are tabulated for a large range of frequencies. Should I calculate a complex permittivity and model that as a multi-debye or is there a way to model permittivity and conductivity separately? 
I looked at the [video](https://www.youtube.com/watch?v=DfHuRxfp7ec&list=PLf_lDHRs1yH6h5T-s-RvPbkdXFMqLrhy4&index=9) of @IraGia and he shows the extended Debye model where conductivity is apart. But it was not clear to me which conductivity to use since it is frequency-dependent.

IraGia:
Hi Alberto,

Electric permittivity is a complex function with both real and imaginary parts. In general, as a hard and fast rule, the real part dictates the velocity of the medium while the imaginary part is related to the electromagnetic losses. Conductivity and permittivity are often separated and described independently (for valid reasons), but both permittivity and conductivity are related to the same property i.e. electromagnetic losses. To that extent, conductivity can be incorporated into a more inclusive complex number that includes the permittivity (both real and imaginary parts) plus the conductivity. If the permittivity is a Debye function, then the inclusive function is called "extended Debye", meaning that is a typical Debye plus a conductive term. The generic form of "extended media" looks like this: 

ε=ε' + jε'' + σ/(jω) 

where ω is the angular frequency, σ is the conductivity (can be frequency dependent) and ε' and ε'' are the real and imaginary parts of the permittivity respectively.  Both the real and imaginary parts can be frequency dependent; and for the extended Debye this equals with 

ε=ε_inf + De/(1+jωt0)+ σ/(jω) 

From the above it is apparent that the conductivity can be implemented as part of the imaginary part of an extended permittivity model. Therefore, you can use multi-Debye functions to fit the real and the imaginary parts (as you would do for a Cole-Cole medium) and implement this type of media in FDTD. 

In gprMax you can do that by using the class that allows you to fit an arbitrary function with multi-Debye (Rawdata). First you have to built this function i.e. implement the frequency dependent conductivity as part of the complex permittivity and save the real and the imaginary part in a file. Subsequently, use this file as input to "Rawdata()" class. 

Hope this was helpful. 

Regards,
Iraklis

alberto-battistel:
Hi Iraklis,
ok, I got it. Basically what I needed to do was to take the real and the imaginary part of the complex dielectric permittivity and pass it to your Debye_fit.Rawdata. Much easier then I thought.
My complex permittivity was defined as:
formula^{\alpha_i}}%20%2B%20\frac{\sigma}{j\omega\epsilon_0})
and I could fit resonably good in the frequency range I am interested with only one or two time constants without any additional conductivity term. Nice!!
I may play around with it in the future to have a direct least-square fit instead of a particle swarming process, but it is already working astonishingly well, thanks.

alberto-battistel:
Hi Iraklis,
I am using the debye fitting for a publication and I would like to cite it. 
Do you have a particular article where it is explained? Will it eventually merge with the main GPRMax repository?

Alberto

IraGia:
Hi Alberto,

 You can just cite the gprMax papers that you can find here (https://www.gprmax.com/publications.shtml). You can also use this paper:
D. F. Kelley, T. J. Destan and R. J. Luebbers, "Debye Function Expansions of Complex Permittivity Using a Hybrid Particle Swarm-Least Squares Optimization Approach," in IEEE Transactions on Antennas and Propagation, vol. 55, no. 7, pp. 1999-2005, July 2007,

I am using a modified version of what Kelley et al. are doing. Their approach is unstable and requires regularisation that is tuned via trial and error. The gprMax script is unconditionally stable without the need for regularisers. 


With respect to using least squares instead of PSO, you can check this conference paper:

I. Giannakis, A. Giannopoulos and N. Davidson, "Incorporating dispersive electrical properties in FDTD GPR models using a general Cole-Cole dispersion function," 2012 14th International Conference on Ground Penetrating Radar (GPR), 2012, pp. 232-236
 
The problem with hybrid least squares is that they are often stack in local minima and need good initialisation. The modified PSO used in the gprMax script is not affected by this and in contrast to Kelley et al. it is always stable.

Hope this was helpful. 

Iraklis

craig-warren:
Closed, as this functionality will soon be integrated into v.4 via GSoC work.

PR #138, opened 2017-12-15
geometry output speed up

jasminium:
Hi Craig. Here's the speed up for the geometry output files. So far I've only worked on .vtp files but should be able to extend the same thing to the vtk image data files too.


craig-warren:
Looks good. Is is possible for you to do the same for the `.vti` files as well? Then we can include both at the same time. 

Also can the `.pyx` file be named `geometry_outputs.pyx` for consistency with other file naming? I think this is possible whilst not conflicting with `geometry_outputs.py`. What do you think?

jasminium:
Thats the `.vti` files too.

There won't be a conflict importing the Cython module from `geometry_outputs.py` but there will be from `model_build_run.py`. There's probably another way to do it using the import module. However, it might be more straightforward to have a naming convention for the extensions? That way if you see `import some_function from module_ext` its clear than it's a Cython extension.

PR #147, opened 2018-03-17
Bugfix for when no rx commands and using GPU

Tagussan:
Snapshot is available with GPU. And bugfix for simulation with GPU when no rx exists.


Tagussan:
@craig-warren Thank you for reviewing my pull request.

Oh, that's why it took long time writing to file with this method... I reverted the commit for snapshot. 
Looking forward to use next version :)

PR #161, opened 2018-06-07
Snapshot speedup cpu

denck007:
Removed all loops from the snapshot file and only using numpy array math to compute values. The data for each field type is saved once instead of one value at a time.

When running the cylinder_Ascan_2D.in file with the snapshot code it took ~10.5 seconds before, now it takes ~1.5 seconds.


jasminium:
Looks great! Nice speed

One thing - is it possible to use a single reshape call instead of transposing each field array?
`
E = np.stack((Ex.transpose(2,1,0),Ey.transpose(2,1,0),Ez.transpose(2,1,0)),axis=-1)
E = np.stack(Ex, Ey, Ez).reshape(-1, order='F')
`

jasminium:
Also added field slicing between xs ... xf etc

craig-warren:
@denck007 @jasminium thanks guys! I have also been working on a Cythonized speedup for snapshot writing on CPU, as well as implementing CUDA-based snapshots for GPU. So I'll incorporate this into it. I still have a bit of work to do to finish it off, as I want to retain different snapshot writing methods (to disk and memory) depending on user requirements/capabilities. I hope to get this finished in the next few weeks.

craig-warren:
@denck007 @jasminium I have overhauled the snapshot writing code in https://github.com/gprMax/gprMax/tree/v.3.1.4 You can now use snapshots with the GPU-based solver, and the performance of snapshot writing with the CPU-based solver should be improved. Give it a try and report any problems. Thanks.

PR #194, opened 2019-03-13
added B-scan plotting options

rjdelf:
Allow plotting of rx_array output in a single b-scan and added agc and geometrical spreading correction when plotting a b-scan.
invoke using an optional --scale 'agc' or 'gsc' after the call to tools.plot_Bscan


craig-warren:
Introducing a `-gather` flag into the plot_Bscan function in the developer branch.

PR #207, opened 2019-05-15
use process_time to measure only process execution time

jasminium:
were currently using perf_counter() to measure execution time. suggesting to we move to process_time(). from the docs process_time() measures only the process wide time where as perf_counter() measure the system wide time. system wide time presumably includes other processes. This is not so clear from the docs? in environments where multiple users are running scripts process_time should provide a more accurate execution time for gprMax scripts.

### DIFF
--- a/gprMax/gprMax.py
+++ b/gprMax/gprMax.py
@@ -24,7 +24,6 @@ import os
 import sys
 
 from enum import Enum
-from time import perf_counter
 
 import h5py
 import numpy as np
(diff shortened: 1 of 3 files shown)


craig-warren:
Sounds sensible. There are also uses of `perf_counter()` in `gprMax.py` which will also need updating.

jasminium:
ah thanks. have replaced all the instances now

PR #210, opened 2019-06-09
opencl feature addition code-base

spaceVStab:
- Written PyOpenCl class which will take care of selecting platforms and devices, initializing contexts and queues and also primary solver. 
- Written kernel for basic updates of E and H fields, output storing and source updates.
- Issues with building with Nvidia platforms.
- Although building, there is no correct output.

### DIFF
--- a/gprMax/gprMax.py
+++ b/gprMax/gprMax.py
@@ -22,6 +22,7 @@ import argparse
 import datetime
 import os
 import sys
+import warnings
 
 from enum import Enum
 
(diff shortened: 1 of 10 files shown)


spaceVStab:
Yes, all the fields components are zero due to same reason maybe. 
I will look into the dev branch and let you know if anything comes up.

craig-warren:
Sorry the branch is named 'config' not 'dev'.

PR #225, opened 2019-08-28
Bug instability

jasminium:
Hello,

news on the mkl bug. removing -fopenmp from the linker args in setup.py appears as a fix. Do we require this for anything else?

### DIFF
--- a/gprMax/gprMax.py
+++ b/gprMax/gprMax.py
@@ -25,12 +25,6 @@ import sys
 
 from enum import Enum
 
-# There is a bug with threading and MKL on macOS
-# (https://github.com/gprMax/gprMax/issues/195) . Setting the MKL threading
-# layer to sequential solves it, but must be done before numpy is imported.
-if sys.platform == 'darwin':
-    os.environ["MKL_THREADING_LAYER"] = 'sequential'
-
 import h5py
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


agianno:
This should be needed for linking the OpenMP libraries. So without there might not be threading support. Let’s double check

jasminium:
checking this now. With and without this option the following GCC call is made. Both have the ``-fopenmp`` flag so I think we should be ok.

`gcc-7 -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/Users/s1563137/miniconda3/envs/gprMaxSummer/include -arch x86_64 -I/Users/s1563137/miniconda3/envs/gprMaxSummer/include -arch x86_64 -I/Users/s1563137/miniconda3/envs/gprMaxSummer/lib/python3.7/site-packages/numpy/core/include -I/Users/s1563137/miniconda3/envs/gprMaxSummer/lib/python3.7/site-packages/numpy/core/include -I/Users/s1563137/miniconda3/envs/gprMaxSummer/include/python3.7m -c gprMax/cython/fields_updates_dispersive.c -o build/temp.macosx-10.9-x86_64-3.7/gprMax/cython/fields_updates_dispersive.o -O3 -w -fopenmp -march=native`

Only reference I can find to ``extra_link_args`` is to compile cython from  [Jupiter notebook](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiling-with-a-jupyter-notebook)

jasminium:
Some more information about this here

https://github.com/ContinuumIO/anaconda-issues/issues/8803

PR #233, opened 2019-12-12
New MPI executor

tschruff:
I added a new generic MPI executor class and accompaying CLI in preparation for tests on JURECA. Unfortunately I also had to change some of the existing code due to some problems, e.g. it is not possible to pickle open file-like objects and therefore its not possible to send them via MPI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 *.pyc
 *.pyo
 
+# Python virtual environment
+venv/
+
 # Cython
 *.c
 *.so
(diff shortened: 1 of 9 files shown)


craig-warren:
@tschruff thanks for this great work. I'm closing this PR as I am working to integrate your MPIExecutor into the new version of gprMax which features sub-gridding and a re-structure of a lot of the code which simplifies things and makes future development easier. It is currently the `config` branch - https://github.com/gprMax/gprMax/tree/config

PR #244, opened 2020-04-14
Geometry file produced only once

zeticabrian:
- Ensure geometry file produced only once if geometry_fixed option specified.
- Enforce underscore before model run number on geometry files (when more than one produced)

### DIFF
--- a/gprMax/model_build_run.py
+++ b/gprMax/model_build_run.py
@@ -104,7 +104,8 @@ def run_model(args, currentmodelrun, modelend, numbermodelruns, inputfile, usern
     global G
 
     # Used for naming geometry and output files
-    appendmodelnumber = '' if numbermodelruns == 1 and not args.task and not args.restart else str(currentmodelrun)
+    appendmodelnumber = '' if numbermodelruns == 1 and not args.task and not args.restart else '_'+str(currentmodelrun)
+    appendmodelnumberGeometry = '' if numbermodelruns == 1 and not args.task and not args.restart or args.geometry_fixed else '_'+str(currentmodelrun)
 
     # Normal model reading/building process; bypassed if geometry information to be reused
     if 'G' not in globals():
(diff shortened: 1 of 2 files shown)


zeticabrian:
I realise now there is more work to do here. The addition of the underscore on multiple file outputs needs to be addressed in the outputfiles_merge function.

craig-warren:
@zeticabrian thanks for investigating this. Our intension is for users to run their model once with the `--geometry-only` flag, and then when they are happy it is OK run the simulation properly (with any other flags such as `--geometry_fixed) having commented out any `#geometry_view` commands in their input file.

zeticabrian:
@craig-warren Noted. 
I am trying to build a team environment with a growing collection of models, created and run using the api() instead of command line calls. These models (once commited to our project repository) ideally dont need editing. This way, users can get the geometry file and the complete model just by executing the code.

Happy to continue using our own fork if this doesn't fit with your philosophy for gprMax.

craig-warren:
Interesting work. The next release of gprMax which we are developing (https://github.com/gprMax/gprMax/tree/config) has a proper API that can be used instead of the hash command based input file. So you will be able to write a pure Python (`.py`) script with everything for your gprMax model.

zeticabrian:
@craig-warren just a follow-up from curiosity... why do gprMax input files use the reverse comment syntax to Python (hashtag for code, and no hashtag for comments)? 
Tempted to reverse this in our fork, but your planned "proper API" might make that effort unecessary.

craig-warren:
@zeticabrian, when @agianno originally created gprMax (in 1996), it was written in C, not Python. We retained the 'hash' commands when I re-wrote gprMax in Python because so much of our user-base was already familiar with them. It is not ideal given they are comments in Python. However, I think you'll like our proper API which allows you to avoid the hash commands altogether if you choose. It basically allows you to interact with gprMax as a Python package, so is more aimed at those with code/programming experience. I still have a bit more work to do on the new release as well as testing but we hope it will be out in the summer.

PR #268, opened 2020-11-05
Fixing bug: pml_updates (gpu) compilation

Keredu:
I have tried to run gprMax from a different directory and I got "import gprMax.pml_updates.pml_updates_electric_HORIPML_gpu" error. In the lines 72-76, we store all the pyx files in a list but, when we go to gprMax/pml_updates we can see that pml_updates_electric_HORIPML_gpu is .py instead of .pyx, so they are not compiled.


craig-warren:
All of the CUDA (GPU) functions should be within standard Python modules (.py). They do not get compiled in the same way as the Cythonized functions (.pyx), so they should not be .pyx files. The CUDA functions are compiled at runtime by the NVCC compiler as part of how PyCuda works.
If you are having an import problem please report it in our issue tracker along with the exact error message and someone will take a look.

PR #305, opened 2022-02-12
Added --save argument for plotting A-scan and B-scan figures

spaceVStab:
While working with gprMax in terminal which do not support graphical interfaces, the figures can be saved using --save argument.


craig-warren:
Added `-save` argument to plot_Ascan and plot_Bscan in devel branch.

PR #314, opened 2022-07-31
conda env for newer python version and cuda support

swenkel:
- update to newer python version and added full support for cuda by default
- original env file is not replaced/changed


craig-warren:
Have successfully checked the devel branch with Python 3.10. Not sure the CUDA tools/libs are needed in the Python environment.

PR #320, opened 2022-09-06
first commit with 2022 400MHz antenna model

mzunguman:
New function to add new 2022 Model of the 400MHz antenna based on Stadler et al.


craig-warren:
@mzunguman can you do a new PR for this into the devel branch please. I'd like to include your 2022 model when we make the devel branch the new master branch which should happen in the next few months.

PR #332, opened 2023-01-16
Stadler new400 m hz antenna model 2022

mzunguman:
New 400 MHz antenna model 2022 by Sam Stadler


craig-warren:
@mzunguman, sorry for the very long delay with this. I'm closing the PR as I've added the updated GSSI 400MHz antenna model to the 'devel' branch of gprMax which will become the main 'master' branch in the next month or so. If you 'pull' the devel branch and check the changes I'd be grateful, thanks.

PR #341, opened 2023-04-15
This PR updates a few files which makes them much cleaner with better code.

Sai-Suraj-27:
While going through the code in a few files I found that in many places the code can be updated to make it more readable and clean. So, I went through a few files and changed some of their code to make them better. **For example: In many places, an f-string could have been used but it was not used, So I used f-strings in those places. In some places multiple comparisons were made using the == operator I replaced those with an in operator, like in the waveforms.py file.** I tried my best to make the code cleaner and look better.


craig-warren (approved):
Thanks for this - agree with most of the changes but some of the formatting style is not consistent with what we use elsewhere so will revert that

Sai-Suraj-27:
Ok. Got it...👍, Thank you.

PR #348, opened 2023-05-24
Add ellipsoid to available geometries

randand:


craig-warren:
@randand thanks for this. We are not looking to wholesale backport new geometries from the devel branch to the current version (3.1.6) so closing this PR.

PR #353, opened 2023-06-23
Refactoring of some code to make it more readable and clean.

Sai-Suraj-27:
# Description
I found that in many places the code can be updated to make it more readable and clean. So, I went through those files and changed some of their code to make them better.

## Type of change
Refactoring of some code to make it better.

# Checklist:
- [x] I have made this from my own
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] The title of my pull request is a short description of the requested changes.

### DIFF
--- a/gprMax/cmds_geometry/add_grass.py
+++ b/gprMax/cmds_geometry/add_grass.py
@@ -111,7 +111,7 @@ class AddGrass(UserObjectGeometry):
             if ys == yf or zs == zf:
                 logger.exception(f'{self.__str__()} dimensions are not specified correctly')
                 raise ValueError
-            if xs != volume.xs and xs != volume.xf:
+            if xs not in [volume.xs and volume.xf]:
                 logger.exception(f'{self.__str__()} must specify external surfaces on a fractal box')
                 raise ValueError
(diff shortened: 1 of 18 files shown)


craig-warren (commented):
@Sai-Suraj-27 can you check the ``or`` condition that has been removed in places in the ``add_grass`` and ``add_surface_roughness`` modules. Specific example at line 133 in ``add_grass``.

Sai-Suraj-27:
> 



> @Sai-Suraj-27 can you check the `or` condition that has been removed in places in the `add_grass` and `add_surface_roughness` modules. Specific example at line 133 in `add_grass`.

@craig-warren. I removed the `or` conditions in the `add_grass` and `add_surface_roughness` modules because we are already checking `if xs==xf:` in the above if condition (at line 110 in `add_grass`), and we are not modifying `xs or xf` anywhere so when we go inside the elif condition `ys==yf` it already means that `xs!=xf`, and when we are inside the `elif zs==zf:` condition both `xs!=xf`, and `ys!=yf`. So, I think it is not necessary to check them again in the `or` conditions.

PR #403, opened 2023-11-14
Fixed grid global object persisting through multiple python api calls in master branch

ThomasRigoni7:
# PR Description

In the master branch, the `FDTDGrid` object G in `model_build_run.py` is not deleted at the end of multiple simulation runs (after multiple A-scans).

This causes the object to be still in the globals() of the module when the gprMax APIs are called again inside the same python program, causing all configuration and geometry to not be recalculated and be incorrectly derived only from the first input file, even in the second run.

## 🛠️ Related Issue (Number)

Closes # 402

## Type of change

## Checklist


ThomasRigoni7:
The change introduces new problems when the 'restart' option is used, see #405

craig-warren:
Superceded by #405

PR #521, opened 2025-11-15
Add --save and --output-dir options for plot_Bscan

Agrimsh:
# PR Description

This PR adds the ability to save B-scan plots directly from tools/plot_Bscan.py using a new --save argument.
Previously, the tool only supported plt.show(), which caused failures in Docker/HPC environments with no display.
This enhancement allows users to generate plots non-interactively.

Closes #

## Type of change

Added --save command-line option to specify output file path.

Added optional --output-dir argument (if you implemented it).

Enabled saving plots using fig.savefig().

Ensured backward compatibility — behaviour is unchanged if --save is not supplied.

The plotting tools currently cannot save figures unless modified manually.
Users running gprMax inside:

Docker containers

HPC clusters

Headless servers

…cannot use plt.show() because it fails without a display.

This PR solves that problem by allowing plots to be saved directly to file.

### DIFF
--- a/tools/plot_Bscan.py
+++ b/tools/plot_Bscan.py
@@ -72,7 +72,7 @@ def mpl_plot(filename, outputdata, dt, rxnumber, rxcomponent):
     # fig.savefig(path + os.sep + savefile + '.png', dpi=150, format='png', 
     #             bbox_inches='tight', pad_inches=0.1)
 
-    return plt
+    return fig
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for adding these plotting options. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. This is a new utility feature rather than a v3 bug fix, so we will not extend master at this stage. Please propose the feature against devel if it is still needed for v4.

Issue #519, opened 2025-11-03
Why don’t the plotting tools support saving plots to file?


Franreno:
When I was running the [Bowtie Antenna Model](https://docs.gprmax.com/en/latest/examples_antennas.html#bowtie-antenna-model) example  inside a Docker container, the plotting tools (`plot_Ascan`, `plot_Bscan`) fails to display plots because they rely solely on `plt.show()`. There’s no command-line option to save figures to file. However, all the scripts already contain commented-out `fig.savefig()` lines.

Is there a specific reason this capability was disabled or never exposed as an option?

Having a `--save` or `--output-dir` option would make these tools far more usable in Docker and HPC environments, removing the need to copy `.out` files outside the container and allowing plots to be generated directly using the gprMax installation that I have inside Docker.

dushyantzz:
Hii @Franreno 
This PR implements the --save option for all main plotting tools (plot_Ascan, plot_Bscan, etc.) so that figures can be saved to a file in headless environments (like Docker, HPC, CI, etc.).
It preserves the existing plt.show( ) behavior when --save isn’t used and adds tests and documentation updates as well.
Happy to make any adjustments based on feedback , looking forward to your review!

pratiktech28:
can i pick solve this issue and assign me with regards

agianno:
Fixed on devel in #732. testing/test_experimental.py/test_models.py (the modern location of the files this issue originally named — they moved out of tests/ in a 2022 restructure) now guard their executable body behind main()/if __name__ == "__main__", a new handle_plot_output() utility auto-saves plots with clear feedback on non-interactive backends across all four plotting tools, and the args.outputs/args.output typo is fixed in the same pass. Closing.

PR #522, opened 2025-11-16
Add save option ascan

Agrimsh:
### Summary

This PR adds a `--save` option to `plot_Ascan.py`, allowing users to save A-scan 
plots directly to a file instead of relying only on `fig.show()`. 
This improves usability in headless environments such as Docker containers, 
HPC clusters, CI pipelines, and remote servers where GUI windows cannot open.

### What Was Changed

- Added `--save` command-line argument.
- Modified the script to call `fig.savefig()` when `--save` is provided.
- Ensured `mpl_plot()` returns a `fig` object for saving.
- Preserved existing behavior (`fig.show()`) when `--save` is not used.
- No changes to simulation or plotting logic—only the output mechanism.

### Why This Is Useful

This brings `plot_Ascan` to parity with the recently updated `plot_Bscan` 
(and addresses the same class of problems raised in Issue #519).
Users working in non-GUI environments can now generate plots without 
needing an interactive display

### DIFF
--- a/tools/plot_Ascan.py
+++ b/tools/plot_Ascan.py
@@ -209,17 +209,24 @@ def mpl_plot(filename, outputs=Rx.defaultoutputs, fft=False):
 
     f.close()
 
-    return plt
+    return fig
 
 
 if __name__ == "__main__":
 
     # Parse command line arguments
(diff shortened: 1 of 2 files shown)


agianno:
Thanks for this — the -save option you're adding here already exists in devel's current plotting tools. This will soon be the default branch and the future development of the code will be on this.

Issue #498, opened 2025-07-22
issue with installation of mpi4py_fft


ofek-presler:
Hello, I've been trying everything to get mpi4py_fft installed, I'm a novice to python and coding so I'd appreciate any help.
I downloaded fftw although I'm not sure I put it where I should have after I've unzipped it, I think I put the files I needed to in the lib and include files in the gprMax env (as well as in the library\lib and library\include) as a solution to my problem but it did not work.
I'm on windows so conda install does not have the package.

when running: pip install mpi4py, it says it cant find the fftw libraries in the lib folder.
I get the error bellow:
error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [17 lines of output]
      Traceback (most recent call last):
        File "C:\Users\276493\anaconda3\envs\gprMax\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 389, in <module>
          main()
        File "C:\Users\276493\anaconda3\envs\gprMax\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\276493\anaconda3\envs\gprMax\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 175, in prepare_metadata_for_build_wheel
          return hook(metadata_directory, config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\276493\AppData\Local\Temp\pip-build-env-yram6s3c\overlay\Lib\site-packages\setuptools\build_meta.py", line 374, in prepare_metadata_for_build_wheel
          self.run_setup()
        File "C:\Users\276493\AppData\Local\Temp\pip-build-env-yram6s3c\overlay\Lib\site-packages\setuptools\build_meta.py", line 317, in run_setup
          exec(code, locals())
        File "<string>", line 248, in <module>
        File "<string>", line 158, in get_extensions
        File "<string>", line 80, in get_fftw_libs
      distutils.errors.DistutilsPlatformError: No FFTW libraries found in ['C:\\Users\\276493\\anaconda3\\envs\\gprMax\\lib']
      [end of output]

[...]

PR #524, opened 2025-11-25
Document that mpi4py_fft is not supported on Windows

Agrimsh:
# PR Description

As discussed in Issue #498 and requested by @craig-warren, mpi4py_fft cannot be
installed on Windows due to the lack of FFTW libraries. This PR adds a clear
documentation note in the Windows installation section to inform users and
avoid installation errors and confusion.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -116,6 +116,18 @@ macOS
 
 Microsoft Windows
 ^^^^^^^^^^^^^^^^^
+.. note::
+
+    The `mpi4py_fft` library is **not supported on Windows**. The required FFTW
+    libraries are not available on Windows by default, and building them manually
+    is non-trivial. As a result, attempts to install `mpi4py_fft` will fail with
+    errors such as:
+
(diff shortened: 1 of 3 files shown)


agianno:
Thanks for this. Both parts of this PR are already covered on devel, independent of this PR: the mpi4py_fft-not-supported-on-Windows note is already in README.rst (see the FFTW/mpi4py_fft install section), and the -save option for plot_Ascan.py/plot_Bscan.py already exists in devel's current plotting tools (in toolboxes/Plotting/, following a 2022 restructure that also moved these files out of tools/).

More generally: master isn't the active development branch here — devel is, and master is currently about 280 commits behind it, so PRs against master end up needing separate review from what's actually shipping. Since both pieces of this one are already present on devel, there isn't a remaining gap to close by retargeting. Closing this — thanks for flagging the Windows install issue originally, and sorry it arrived a bit late relative to devel's own fixes.

PR #532, opened 2025-12-28
Added a faster one-sided FFT-power function for real valued signals.

VyasGuru:
## Summary

This PR adds a new helper function `fft_power_real` that computes a
one-sided FFT power spectrum for real-valued signals using `np.fft.rfft`.

Compared to the existing `fft_power`, this implementation:
- is ~2–3× faster for real-valued inputs
- avoids computing redundant negative-frequency components
- reduces memory usage

## Benchmarks

Measured using representative signal sizes and both power-of-two and
non–power-of-two lengths.

| Condition        | Dtype    | Orig (ms) | Opt (ms) | Speedup |
|------------------|----------|-----------|----------|---------|
| Power of 2       | float32  | 5.2890    | 1.9782   | 2.67×   |
| Power of 2       | float64  | 5.5249    | 1.9554   | 2.83×   |
| Non-Power of 2   | float32  | 9.9420    | 5.3291   | 1.87×   |
| Non-Power of 2   | float64  | 10.3145   | 5.3858   | 1.92×   |

## Files Modified
- `gprMax/utilities.py`

### DIFF
--- a/gprMax/utilities.py
+++ b/gprMax/utilities.py
@@ -142,6 +142,7 @@ def round32(value):
 
 
 def fft_power(waveform, dt):
+
     """Calculate a FFT of the given waveform of amplitude values;
         converted to decibels and shifted so that maximum power is 0dB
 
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for the contribution. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. This PR adds a new FFT utility rather than correcting a critical v3 bug, and the helper is not integrated into an existing workflow. We are therefore closing it; a v4 version can be proposed against devel if there is a current use case.

PR #533, opened 2025-12-28
Optimized rounding utilities by replacing Decimal and ceil

VyasGuru:
## Summary

This PR optimizes the rounding utilities by replacing `decimal.Decimal`-based
rounding and `np.ceil`-based alignment with faster native arithmetic and
bitwise operations.

The changes preserve the existing rounding semantics while reducing overhead
in performance-critical paths.

## What Changed

- **`round_value`**
  - Replaced `decimal.Decimal.quantize` with native arithmetic
  - Preserves `ROUND_HALF_DOWN` behavior for integer rounding
  - Preserves `ROUND_FLOOR` behavior for fractional rounding
- **`round32`**
  - Replaced `np.ceil`-based computation with a bitwise alignment operation

## Benchmarks

Benchmarks were run using freshly generated random input on each iteration to
avoid caching or constant-folding effects.

| Function        | Original (s) | Optimized (s) | Speedup |
|-----------------|--------------|---------------|---------|
| round_value(0)  | 0.7664       | 0.6303        | 22%     |
| round_value(3)  | 1.0078       | 0.6509        | 55%     |
| round32         | 0.6279       | 0.4479        | 40%     |

## Impact

- **Performance**: ~20–55% speedup for rounding utilities depending on mode
- **Code quality**: Removes reliance on `decimal.Decimal` in hot paths
- **Behavior**: No API changes; rounding semantics preserved for supported inputs

## Files Modified

- `gprMax/utilities.py`

### DIFF
--- a/gprMax/utilities.py
+++ b/gprMax/utilities.py
@@ -126,19 +126,23 @@ def round_value(value, decimalplaces=0):
 
     # Rounds to nearest integer (half values are rounded downwards)
     if decimalplaces == 0:
-        rounded = int(d.Decimal(value).quantize(d.Decimal('1'), rounding=d.ROUND_HALF_DOWN))
+        if value >= 0:
+            rounded = int(np.ceil(value - 0.5))
+        else:
+            rounded = int(np.floor(value + 0.5))
 
     # Rounds down to nearest float represented by number of decimal places
     else:
-        precision = '1.{places}'.format(places='0' * decimalplaces)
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for the proposed optimisation. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. Changing the rounding implementation could affect cell-index calculations near rounding boundaries and would require targeted regression testing. This is not an essential v3 bug fix, so we are closing the PR rather than changing the legacy branch. Any still-relevant optimisation should be proposed against devel or the new gprMax master once it is released.

PR #537, opened 2026-01-23
Align source waveform updates with integer time steps

Nesar976:
This PR updates the source waveform sampling logic to align voltage source updates with integer (whole) time steps.

Previously, waveform sampling used half-step offsets, which could introduce a time-staggering mismatch between source updates and the E-field update loop. Aligning these updates to integer time steps improves consistency and avoids unintended phase offsets in simulations.

The change is intentionally scoped to `sources.py` and does not modify behavior outside source timing logic.

### DIFF
--- a/gprMax/sources.py
+++ b/gprMax/sources.py
@@ -101,7 +101,7 @@ class VoltageSource(Source):
                                     * self.waveformvalues_wholestep[iteration] 
                                     * (1 / (self.resistance * G.dy * G.dz)))
                 else:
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for investigating the source timing. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. This PR changes hard voltage-source timing but does not cover the complete source-staggering model, including resistive voltage and magnetic sources, and it has no numerical parity regression tests. The source implementation has been addressed more comprehensively in v4, so we will not alter v3 numerical results with this partial change.

PR #538, opened 2026-01-23
Add Taginfo project file for World-Map-Explorer

Nesar976:
This PR adds a taginfo.json file so the OpenStreetMap community can see which OSM tags are used by World-Map-Explorer.

It includes basic project metadata and the confirmed tags referenced by the app, along with a short README update explaining the file. This is a documentation/metadata change only and does not affect runtime behavior.


agianno:
We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. The PR title does not match the submitted changes, and its head is the same source-timing commit submitted in PR #537. That change is incomplete relative to the full v4 source-staggering model and would alter v3 numerical results without parity tests. We are therefore closing this duplicate PR.

PR #548, opened 2026-02-02
Fix #plate to apply only electric material properties

### LINKED ISSUE #158: Magnetic material properties & plate objects
We recently noticed a bug in the code concerning the setting of magnetic material properties for `#plate` objects. When a plate is specified the electric material properties of the `#material` used are assigned to electric field components that make up the plate, and the magnetic material properties were assigned to incorrect magnetic field components.

Some thought needs to be given on the best design for the code to allow users to separately specify setting material properties for electric and magnetic field components for `#edge` and `#plate` objects. For example, as we have seen there is no clear method for setting *both* electric and magnetic properties on a `#plate` object. Perhaps separate `#plate_electric` and `#plate_magnetic` commands, or a flag on the existing command to indicate electric or magnetic building. This is something we will look into.

This bug would have affected anyone combining a materials with magnetic properties and the plate command. 

For now, we have updated the code (https://github.com/gprMax/gprMax/commit/c13fbb23142e0c0252ba8ede18406772468bccb6#diff-0c31f357656dc5640678b6a63583dee9) to remove the setting of magnetic properties for the plate command (only electric properties are set), which makes its behaviour consistent with the `#edge` command which only sets electric material properties. 

**N.B: No volumetric (3D) objects are affected by this issue.**

pxxad:
### Summary
This PR fixes the handling of material properties for `#plate` objects.

### Details
- `#plate` objects now apply **only electric material properties**
- Magnetic material assignment is explicitly disabled
- Behaviour is now consistent with the `#edge` command

This addresses issue #158, where magnetic material properties were previously applied to incorrect H-field components for plate objects.

### Notes
- Backward compatible
- No impact on volumetric (3D) objects

Fixes #158

### DIFF
--- a/gprMax/input_cmds_geometry.py
+++ b/gprMax/input_cmds_geometry.py
@@ -277,7 +277,11 @@ def process_geometrycmds(geometry, G):
 
                 for j in range(ys, yf):
                     for k in range(zs, zf):
-                        build_face_yz(xs, j, k, numIDy, numIDz, G.rigidE, G.rigidH, G.ID)
+                        # NOTE:
+                        # Plates only apply electric material properties.
+                        # Magnetic material assignment is intentionally disabled to match #edge behaviour.
+
+                        build_face_yz(xs, j, k, numIDy, numIDz, G.rigidE, None, G.ID)
 
             # xz-plane plate
             elif ys == yf:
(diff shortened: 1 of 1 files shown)


agianno:
Closing as already resolved. The underlying issue this targets (#158) was fixed via commit c13fbb23142e0c0252ba8ede18406772468bccb6, which removed magnetic-property assignment from the #plate command — this predates this PR and is already present in master's history.

Confirming directly against the current code: build_face_yz/build_face_xz/build_face_xy (geometry_primitives_ext) accept a rigidH argument in their signature, but the function bodies never use it — only the electric (Ey/Ez etc.) rigid arrays and ID slots are written. So passing G.rigidH at the call sites today has no effect on plate behavior; this PR's change would be a no-op if merged. #plate is already electric-only, consistent with #edge.

Thanks for the contribution, but no further action needed here — issue #158 is already closed.

PR #560, opened 2026-02-18
ensure correct core count

AdityaShome:
# PR Description
Fixed a `TypeError` in `gprMax/utilities.py` that occurs on Linux systems where the `lscpu` command fails to report "Core(s) per socket".

### The Issue
When `lscpu` output is incomplete or parsing fails, the variable `corespersocket` remains `None`. This causes a crash during the calculation of `physicalcores`:
``` bash
physicalcores = sockets * corespersocket
TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```
changes:
- Initialized variables (sockets=1, threads=1, cores=None) before lscpu parsing to ensure predictable starting state.
- Wrapped the CPU core calculation in a conditional check to identify when detection fails.
- Added a robust fallback to psutil.cpu_count() to derive physical and logical cores when lscpu output is incomplete.

Benefits:
- [x] Prevents the TypeError: unsupported operand type application crash caused by multiplying integer 1 by None.
- [x] Ensures valid integer values for core counts are always returned, allowing simulations to proceed safely even on systems with non standard lscpu output.
- [x] Reliable CPU detection in constrained environments where lscpu might be missing and cross platform compatibility.

Closes #

## Type of change

## Checklist

- [x] I have added comments for my code, particularly in hard-to-understand areas.

### DIFF
--- a/gprMax/utilities.py
+++ b/gprMax/utilities.py
@@ -295,6 +295,10 @@ def get_host_info():
             for line in cpuIDinfo.split('\n'):
                 if re.search('model name', line):
                     cpuID = re.sub('.*model name.*:', '', line, 1).strip()
+            
+            sockets = 1
+            threadspercore = 1
+            corespersocket = None
             allcpuinfo = subprocess.check_output("lscpu", shell=True, env=my_env, stderr=subprocess.STDOUT).decode('utf-8').strip()
             for line in allcpuinfo.split('\n'):
                 if 'Socket(s)' in line:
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for investigating core-count detection. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. The submitted change does not catch the ValueError raised when lscpu reports a non-numeric socket or core value, so the reported failure can still occur before the fallback is reached. Since the fix is incomplete and v4 has newer detection logic, we are closing the PR.

PR #566, opened 2026-02-22
Add --save option to plotting tools for headless/Docker environments

### LINKED ISSUE #519: Why don’t the plotting tools support saving plots to file?

dushyantzz:
Add a --save <output_file> CLI option to all main plotting tools (plot_Ascan, plot_Bscan, plot_antenna_params, plot_source_wave). When --save is used, figures are written to the given path with fig.savefig() and parent directories are created if needed; when it is not used, behavior is unchanged and plt.show() is still used.
Motivation and context: Plotting tools currently rely only on plt.show(), which fails in headless environments (Docker, HPC, SSH). There was no way to save plots from the command line, even though savefig was present in comments. This change makes the tools usable in CI, containers, and clusters and matches the use case described in the related issue.

Changes:
CLI: New optional argument --save <output_file> on all four plotting tools.
Behavior: If --save is set: save to file (creating dirs with os.makedirs(..., exist_ok=True)); if not set: keep existing plt.show() behavior.
Naming: Multiple receivers get _rxN before the extension (A-scan/B-scan); antenna params produce _tl_params and _ant_params from the given base path.
Tests: tests/test_plot_save.py added; two tests confirm that plot_source_wave and plot_Ascan create a non-empty file when --save is used; the B-scan test is skipped when no merged B-scan output file exists.
Docs: docs/source/plotting.rst and README.rst updated with --save usage and examples.

Closes #519 

Type of change
[ ] Bug fix (non-breaking change which fixes an issue)
[x] New feature (non-breaking change which adds functionality)
[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[x] This change requires a documentation update.
Checklist
[ ] Did pre-commit passed all checks?
[x] I have performed a self-review of my code.
[x] I have added comments for my code, particularly in hard-to-understand areas.
[x] I have made corresponding changes to the documentation.
[x] My changes generate no new warnings.
[x] The title of my pull request is a short description of my changes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -161,6 +161,12 @@ When the simulation is complete you can plot the A-scan using:
 
     (gprMax)$ python -m tools.plot_Ascan user_models/cylinder_Ascan_2D.out
 
+To save the plot to a file instead of displaying it (e.g. in Docker or on a headless server), use ``--save``:
+
+.. code-block:: bash
+
+    (gprMax)$ python -m tools.plot_Ascan user_models/cylinder_Ascan_2D.out --save output_plot.png
+
 Your results should be like those from the A-scan from the metal cylinder example in `introductory/basic 2D models section <http://docs.gprmax.com/en/latest/examples_simple_2D.html#view-the-results>`_
 
 When you are finished using gprMax, the conda environment can be deactivated using :code:`conda deactivate`.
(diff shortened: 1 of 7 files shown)


agianno:
Thanks for this — the -save option you're adding here already exists in devel's current plotting tools. This is going to be the default distribution of gprMax and future development will be on it only.

PR #568, opened 2026-02-22
Fix: remove self-dependency on gprMax from requirements.txt

lokeshsoni1:
The requirements.txt file listed gprMax as a dependency.
Since gprMax is installed from source (pip install -e .) and is not
available on PyPI, this caused installation to fail for new users.
This PR removes the self-dependency to allow clean installation.


agianno:
Thank you for identifying the self-dependency problem. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. The requirements cleanup has already been completed by merged PR #585, so this PR is now superseded and can be closed.

PR #569, opened 2026-02-23
Fix memory accumulation in multiscene runs

sahilshingate01:
## Problem

When running multiple scenes using `gprMax.run(scenes=scenes, n=len(scenes))`, 
memory from each scene was not being released after it completed. Each scene 
adds ~2GB to RAM and never frees it, so after enough scenes the process 
crashes with an out-of-memory error.

## Root Cause

After debugging I found 6 separate issues all contributing to this:

- `FDTDGrid` had no teardown method, so large numpy arrays like `Ex`, `Ey`, 
  `Ez`, `Hx`, `Hy`, `Hz`, `solid`, `rigidE`, `rigidH`, `ID`, 
  `updatecoeffsE`, `updatecoeffsH` stayed in memory after each scene
- `Material.maxpoles` is a class-level variable that was never reset, so 
  the dispersive pole count from one scene leaked into the next
- `Snapshot` class variables (`nx_max`, `ny_max`, `nz_max`, `bpg`) had 
  the same problem
- `h5py.File()` in `fields_outputs.py` was not using a context manager, 
  leaving file handles open
- Same file handle leak existed in `geometry_outputs.py` for both the 
  HDF5 and materials text file
- `model_build_run.py` was calling `del G` without first releasing the 
  large arrays `G` was holding

## Changes

- `gprMax/grid.py` — added `FDTDGrid.cleanup()` that explicitly deletes 
  all large field and geometry arrays and clears object lists
- `gprMax/materials.py` — added `Material.reset_class_state()` to reset 
  `maxpoles` between runs
- `gprMax/snapshots.py` — added `Snapshot.reset_class_state()` to reset 
  class-level dimension variables
- `gprMax/fields_outputs.py` — wrapped `h5py.File()` in a `with` block
- `gprMax/geometry_outputs.py` — same context manager fix for both file handles
- `gprMax/model_build_run.py` — call `cleanup()` and both `reset_class_state()` 
  methods before `del G`
- `tests/test_memory_cleanup.py` — 11 unit tests covering all the above

## Testing
```bash
conda activate gprMax
python -m pytest tests/test_memory_cleanup.py -v
```

## Notes

`geometry_fixed` mode is fully preserved — cleanup only runs when geometry 
is not being reused between models.

Fixes #[paste the issue number here]

### DIFF
--- a/gprMax/fields_outputs.py
+++ b/gprMax/fields_outputs.py
@@ -116,46 +116,47 @@ def write_hdf5_outputfile(outputfile, G):
         G (class): Grid class instance - holds essential parameters describing the model.
     """
 
-    f = h5py.File(outputfile, 'w')
-    f.attrs['gprMax'] = __version__
-    f.attrs['Title'] = G.title
-    f.attrs['Iterations'] = G.iterations
-    f.attrs['nx_ny_nz'] = (G.nx, G.ny, G.nz)
-    f.attrs['dx_dy_dz'] = (G.dx, G.dy, G.dz)
-    f.attrs['dt'] = G.dt
-    nsrc = len(G.voltagesources + G.hertziandipoles + G.magneticdipoles + G.transmissionlines)
-    f.attrs['nsrc'] = nsrc
-    f.attrs['nrx'] = len(G.rxs)
-    f.attrs['srcsteps'] = G.srcsteps
-    f.attrs['rxsteps'] = G.rxsteps
+    with h5py.File(outputfile, 'w') as f:
+        f.attrs['gprMax'] = __version__
+        f.attrs['Title'] = G.title
+        f.attrs['Iterations'] = G.iterations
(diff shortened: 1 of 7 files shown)


sahilshingate01:
@craig-warren 
Hi team! I've opened a PR that fixes the memory accumulation issue 
in multiscene runs (#569). Would love if someone could take 
a look when they get a chance. Happy to make any changes based on feedback!

PR link: https://github.com/gprMax/gprMax/pull/569

agianno:
Thank you for the detailed work on multiscene cleanup. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. This PR changes grid cleanup, output ownership, HDF5 handling and several class-level states across the legacy runtime. That is too broad a lifecycle change to merge immediately before archival, particularly because v4 now has a different model lifecycle. We are therefore closing this PR.

PR #578, opened 2026-02-27
Fix receiver start index detection in antenna pattern initial_save

### LINKED ISSUE #577: Antenna Pattern with Antenna_like_GSSI_1500
Hello, 
I am learning how to use the antenna pattern package so I tried to run the example proposed in the gprMax doc (https://docs.gprmax.com/en/devel/inc_AntennaPatterns.html) with the GSSI antenna and the results are... fun. 
Any help on it would be very appreciated. 

Adding informations 
-> I am using the devel version
-> I had to correct correct a line in the plot_fields.py file as it tried to find one more Ex that didn't exist 
correction line 128 : added a "-1" after "start" : (path = "/rxs/rx" + str(rx + start-1) + "/")
-> my command prompt : 
(gprMax-devel) C:\stop\being\curious>python -m toolboxes.AntennaPatterns.plot_fields C:\same\question\antenna_like_GSSI_1500_patterns.npy
Centre frequency: 1.5 GHz
Critical angle for Er 5 is 27.0 degrees
Wavelength: 0.089 m
Observation distance(s) from 0.100 m (1.1 wavelengths) to 0.300 m (3.4 wavelengths)
Theoretical boundary between reactive & radiating near-field (0.62*sqrt((D^3/wavelength): 0.030 m
Theoretical boundary between radiating near-field & far-field (2*D^2/wavelength): 0.081 m

Thank you very much and a great day to anyone reading my issue. 

Image

mohanakatari119-bit:
# PR Description

The `initial_save.py` script hardcoded `start = 2` when `antenna = True`, assuming the antenna's internal receiver (`rxbowtie`) is always the first receiver (rx1) in the output file. This assumption holds when the antenna function is called before the observation receivers in the input file, which is the case in the example `.in` files. However, if a user places the observation `#rx:` points before the antenna call, `rxbowtie` ends up as the last receiver. The loop then attempts to read `Ex`, `Ey`, `Ez`, `Hx`, `Hy`, `Hz` from `rxbowtie`, which only saves a single field component (`Ey` or `Ex`), causing a `KeyError`.

The fix dynamically determines the correct `start` index by inspecting the first receiver's stored field components. If the first receiver has all six default outputs (`Ex`, `Ey`, `Ez`, `Hx`, `Hy`, `Hz`), it is an observation receiver and `start` is set to `1`. If it only has a subset (as is the case for `rxbowtie`), it is the antenna's internal receiver and `start` is set to `2`. This makes the code robust to both receiver orderings.

## 🛠️ Related Issue (Number)

Closes #577

## Type of change

## Checklist

- [x] I have added comments for my code, particularly in hard-to-understand areas.

### DIFF
--- a/user_libs/antenna_patterns/initial_save.py
+++ b/user_libs/antenna_patterns/initial_save.py
@@ -77,8 +77,17 @@ iterations = f.attrs['Iterations']
 dt = f.attrs['dt']
 nrx = f.attrs['nrx']
 if antenna:
-    nrx = nrx - 1  # Ignore first receiver point with full antenna model
-    start = 2
+    nrx = nrx - 1  # Ignore the antenna's internal receiver point
+    # Determine start index by checking whether the first receiver has all
+    # default field outputs. If it does, the antenna's internal receiver (which
+    # only saves a single component, e.g. 'Ey') must be at the end of the list,
+    # so observation receivers begin at rx1. Otherwise the antenna receiver is
+    # first and observation receivers begin at rx2.
+    first_rx_outputs = list(f['/rxs/rx1/'].keys())
+    if all(c in first_rx_outputs for c in ['Ex', 'Ey', 'Ez', 'Hx', 'Hy', 'Hz']):
+        start = 1  # Antenna receiver is last; observation receivers start at rx1
+    else:
+        start = 2  # Antenna receiver is first; observation receivers start at rx2
 else:
     start = 1
 time = np.arange(0, dt * iterations, dt)


Perpolita:
Do these changes work for you ? bc then for me it searches for a "rx0" that doesn't exist.

Perpolita:
okay my bad your code does run however it still doesn't work lol 
(but it's slightly different from the first one I got)
results_GSSI_pattern_reddit_v2

mohanakatari119-bit:
Thanks for testing this! The indexing fix ensures the right receivers are read, but the chaotic pattern suggests there may be a separate issue with the coordinate transformation or field data mapping.

To help debug further, could you share:

1. **Which `.in` file are you using?** One of the examples (`antenna_like_GSSI_1500_patterns_E.in` / `_H.in`) or a custom one?
2. **In your input file, is the `antenna_like_GSSI_1500(...)` call placed before or after the observation `#rx:` points?** This matters for how the receiver start index is determined.
3. **What are the values in your `_rxsorigin.txt` file?** If this does not match the actual antenna position in the simulation, the relative coordinates used for the spherical transform will be wrong, which would produce exactly the kind of chaotic pattern seen in your plot.
4. **Are you running the H-plane or E-plane case?** The `type` variable in `initial_save.py` (line 29) must match what your simulation computed.

The pattern being "slightly different from the first one" suggests the indexing change did affect which receivers are read, but the resulting coordinates may still not be mapping correctly to the expected radii and angles.

SarahTlr:
Hi! I'm working on this with @Perpolita:
1. We are directly using [examples\antenna_like_GSSI_1500_patterns.py](https://github.com/gprMax/gprMax/blob/devel/examples/antenna_like_GSSI_1500_patterns.py) (no .in files as we are using the devel version)
2. `antenna_like_GSSI_1500(...)` is called before `gprMax.Rx` (we have not modified anything in the file available in examples)
3. The values in `_rxsorigin.txt` seems correct as it matches the antenna position in the simulation (0.125, 0.34, 0.34)
4. We are running the E-plane case, so the initial parameters of `initial_save.py` seems ok (`type="E"`)

Thank you for you reply!

agianno:
Thank you for investigating the antenna-pattern receiver selection. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. The proposed detection assumes that receiver 1 must contain all six field components; a valid receiver saving only selected components would therefore be incorrectly skipped. Since this heuristic is unsafe and the antenna-pattern tooling has changed for v4, we are closing the PR.

PR #588, opened 2026-03-07
Fix issue #158: plate command incorrectly setting magnetic material p…

### LINKED ISSUE #158: Magnetic material properties & plate objects

kallal79:
The #plate command was calling build_face_yz/xz/xy with G.rigidH, causing magnetic material properties to be set for incorrect H-field components.

Changes:
- Remove unused 'rigidH' parameter from build_face_yz, build_face_xz, and build_face_xy in geometry_primitives_ext.pyx. The parameter was never used in the function bodies. Also update all 9 internal call sites within the same file (used by build_triangle and build_cylindrical_sector).
- Update input_cmds_geometry.py: remove G.rigidH from the three plate call sites so plates only set electric material properties, matching #edge behaviour as described in issue #158.
- Fix pre-existing bug in input_cmd_funcs.py: rx() was joining multiple output field components without a space separator (e.g. 'ExEz' instead of 'Ex Ez').
- Add tests/test_plate_materials.py: 43 unit tests covering all three plate orientations (yz, xz, xy), verifying that only electric IDs are set and magnetic IDs are never modified. Tests also verify the new 7-argument API rejects the old 8-argument form.

Note: PR #548 proposed passing None for rigidH, but Cython typed memoryviews cannot accept None at runtime (TypeError). The correct fix is to remove the parameter entirely from the function signatures.

Fixes #158


DtDOW:
you have made changes in master branch, in order for PR to get accepted you need to use devl branch

kallal79:
ok

agianno:
Thank you for investigating the plate material behaviour. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. The submitted branch differs from current master by hundreds of unrelated removals and cannot be merged safely. In addition, the removed rigidH argument is unused by the existing face-building functions, so removing it does not produce the claimed runtime physics correction. We are therefore closing this PR.

kallal79:
OK

PR #590, opened 2026-03-08
Fix unsupported waveform membership checks in inputfile_old2new

### LINKED ISSUE #589: Incorrect use of `is` instead of `in` when checking waveform types in inputfile_old2new.py
While exploring the gprMax codebase I noticed a potential logical issue in:

gprMax/tools/inputfile_old2new.py

In several locations (around lines 334, 355, 376, and 397), the code checks whether a parsed parameter exists in a list of unsupported waveform types (`badwaveforms`). However, the identity operator `is` appears to be used instead of the membership operator `in`.

Example pattern observed:

badwaveforms = ['gaussiandot', 'gaussiandotdot']

if params[3] is badwaveforms:
    raise CmdInputError(...)

Using `is` compares object identity rather than list membership and will always evaluate to False in this situation. As a result, unsupported waveform parameters may not trigger the expected exception.

Suggested change:

if params[3] in badwaveforms:
    raise CmdInputError(...)

This change would correctly check whether the parsed waveform parameter belongs to the list of unsupported waveform types.

If this behavior is unintended, I would be happy to submit a pull request with the fix.

twnkl2713:
# PR Description

This PR fixes a logical issue in `tools/inputfile_old2new.py` where unsupported waveform types were incorrectly checked using the identity operator (`is`) instead of the membership operator (`in`).

In several locations in the file, the code checks whether a parsed waveform parameter belongs to a list of unsupported waveform types:

```python
badwaveforms = ['gaussiandot', 'gaussiandotdot']
```

However, the code used the following condition:

```python
if params[3] is badwaveforms:
```

Since `params[3]` is a string and `badwaveforms` is a list, the identity comparison always evaluates to `False`. As a result, unsupported waveform types such as `gaussiandot` and `gaussiandotdot` would not trigger the expected exception.

This PR replaces the identity comparison with a membership check:

```python
if params[3] in badwaveforms:
```

This ensures that unsupported waveform parameters are correctly detected and handled.

Additionally, the error message formatting was improved by replacing:

```python
''.join(badwaveforms)
```

with

```python
', '.join(badwaveforms)
```

so that unsupported waveform names are displayed clearly in the exception message.

## 🛠️ Related Issue (Number)

Closes #589

## Type of change

## Checklist

### DIFF
--- a/tools/inputfile_old2new.py
+++ b/tools/inputfile_old2new.py
@@ -331,8 +331,8 @@ while(lindex < len(inputlines)):
 # Convert separate #line_source and associated #tx to #waveform and #hertzian_dipole
 for source in linesources:
     params = source.split()
-    if params[3] is badwaveforms:
-        raise CmdInputError("Waveform types {} are not compatible between new and old versions of gprMax.".format(''.join(badwaveforms)))
+    if params[3] in badwaveforms:
+        raise CmdInputError("Waveform types {} are not compatible between new and old versions of gprMax.".format(', '.join(badwaveforms)))
     elif params[3] == 'ricker':
         params[3] = 'gaussiandotnorm'
     waveform = '#waveform: {} {} {} {}'.format(params[3], params[1], params[2], params[4])


agianno:
Thank you for identifying the incorrect waveform membership test. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. The converter contains four occurrences of this error, but the PR corrects only one, leaving the same bug in the Hertzian-dipole, voltage-source and transmission-line paths. Since the submitted fix is incomplete and the converter is legacy v3 tooling, we are closing the PR.

Issue #582, opened 2026-03-04
Memory coalescing improvment  in gprmax/source_updates_gpu.py


Codebruh-sudo:
I was simply going through this  kernel  when I noticed that `srcinfo1` is stored in an Array of Structures (AoS) layout. This causes strided memory access across threads in a warp, which could be reducing memory throughput.
i = srcinfo1[INDEX2D_SRCINFO(src,0)];
j = srcinfo1[INDEX2D_SRCINFO(src,1)];
k = srcinfo1[INDEX2D_SRCINFO(src,2)];
polarisation = srcinfo1[INDEX2D_SRCINFO(src,3)];
From a CUDA performance perspective, this looks like it would cause strided, non-coalesced memory access across the warp. you can tell me if am wrong .i would be happy to hear if there is a reason this layout was chosen. And if this is worth fixing, I can work on it .

SouravVerma-art:
Hi,
I've opened a PR for this issue. Here's a quick summary of what I did:
The root cause was that `srcinfo1` and `srcwaves` were stored in AoS layout, causing strided non-coalesced memory access across warp threads. I converted both to SoA layout and updated the `INDEX2D_SRCINFO` and `INDEX2D_SRCWAVES` macros and all three kernels accordingly.
One thing I want to flag for review: since `NSOURCES` is now baked into the macro at compile time and differs per source type, I compiled a separate `SourceModule` for each source type rather than sharing one. Happy to discuss if there's a cleaner approach. I'm also happy to make any changes based on your feedback.

Thanks for your time.

Codebruh-sudo:
hey , its  nice to see that you have raised the pr . i have shared some technical feedback and a couple of small optimization suggestions over on the PR .

SouravVerma-art:
Hi , thanks for the review and the detailed technical feedback , please let me know if there’s anything else I should improve.

agianno:
Thanks for reviewing the source-array layout.

The proposed coalescing improvement assumes that multiple source threads in the same CUDA warp concurrently load the same member from adjacent source records. That is not how the current v4 `devel` source kernels are launched: they use one thread per block. With only one active source lane in each block, changing `srcinfo1` or `srcwaves` from AoS to SoA cannot provide the claimed intra-warp memory coalescing benefit.

[...]

PR #596, opened 2026-03-14
Fix/srcinfo soa coalesced memory

SouravVerma-art:
# PR Description

## 🛠️ Related Issue #582

Closes #

## Type of change

## Checklist

### DIFF
--- a/gprMax/model_build_run.py
+++ b/gprMax/model_build_run.py
@@ -610,6 +610,7 @@ def solve_gpu(currentmodelrun, modelend, G):
                                        np.int32(snap.yf), np.int32(snap.zs),
                                        np.int32(snap.zf), np.int32(snap.dx),
                                        np.int32(snap.dy), np.int32(snap.dz),
+                                       np.int32(snap.nx), np.int32(snap.ny), np.int32(snap.nz),
                                        G.Ex_gpu.gpudata, G.Ey_gpu.gpudata, G.Ez_gpu.gpudata,
                                        G.Hx_gpu.gpudata, G.Hy_gpu.gpudata, G.Hz_gpu.gpudata,
                                        snapEx_gpu.gpudata, snapEy_gpu.gpudata, snapEz_gpu.gpudata,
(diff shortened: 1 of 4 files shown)


Codebruh-sudo:
hey ,  its good you raised the pr . i saw the improved code nice work on the SoA fix. the macro and the host-side initialisation in gpu_initialise_src_arrays line up correctly so no transpose issues there.
One small thing I noticed. updatecoeffsE and updatecoeffsH are being copied to GPU three times, once for each source type. Since they're the same arrays every time, yoou could just copy them once before the source type checks. the change is tiny   but avoids unnecessary host-to-device transfers.
and also  if a model has all three source types, this triggers three separate JIT compilations at startup because each SourceModule has its own NSOURCES baked in. its fine for the basic models but can cause problems for complex ones. you can correct me if iam wrong .

SouravVerma-art:
you're right about updatecoeffsE and updatecoeffsH, since the arrays are identical for each source type, moving the copy outside the source-type checks makes sense and would avoid redundant host-to-device transfers. I’ll update the implementation accordingly. Regarding the JIT compilation, my understanding was that because each SourceModule  compiles with its own NSOURCES value, models containing multiple source types could trigger multiple compilations at startup. I mentioned it mainly as an observation for complex models, but please correct me if my understanding is inaccurate.

Thanks again for the helpful suggestions.

Codebruh-sudo:
you are right about this bruh .each SourceModule call triggers a separate JIT compilation because NSOURCES is baked in as a compile-time constant via the template substitution. So a model with all three source types compiles three separate kernels at startup. For most typical models this maybe  fine .Thanks for confirming this . i  have already said that  It only becomes noticeable with complex scenes that have many source types or when the kernel template is large.  i was just concerned about the startup time becoming  down the line . by the way great job.

SouravVerma-art:
Hi, I’ve implemented the suggested changes (moved the coefficient copies outside the source-type checks). When you get a chance, could you please take a look? Thanks!

PR #605, opened 2026-03-16
Fix three bugs in setup.py

alphaleporus:
Fix three Python 3 correctness bugs in setup.py that affect error handling and user-facing messages.

## Related Issue (Number)
No existing issue — bugs identified during code review.

## Type of change

## Summary of changes

**1. Missing f-string prefix in version check exit message**
The error message printed literally `{MIN_PYTHON_VERSION[0]}.{MIN_PYTHON_VERSION[1]}`
instead of `3.7` when Python version was too old.

**2. `raise` with a plain string on macOS gcc-not-found path**
Python 3 cannot raise a plain string — doing so raises `TypeError: exceptions must
derive from BaseException` instead of showing the intended message.
Fixed by using `raise SystemExit(...)`.

**3. Bare `except:` in macOS environment variable cleanup**
Only `KeyError` is expected here. Bare `except:` silently swallows
`KeyboardInterrupt` and `SystemExit`. Replaced with `except KeyError:`.
Also simplified the try block by removing the redundant key lookup before deletion.

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ from setuptools import Extension, find_packages, setup
 MIN_PYTHON_VERSION = (3, 7)
 if sys.version_info[:2] < MIN_PYTHON_VERSION:
     sys.exit(
-        "\nExited: Requires Python {MIN_PYTHON_VERSION[0]}.{MIN_PYTHON_VERSION[1]} or newer!\n"
+        f"\nExited: Requires Python {MIN_PYTHON_VERSION[0]}.{MIN_PYTHON_VERSION[1]} or newer!\n"
     )
 
 # Importing gprMax _version__.py before building can cause issues.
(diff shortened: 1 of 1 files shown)


alphaleporus:
Happy to rebase or split this if the format doesn't fit the contribution workflow. Let me know.

agianno:
Thank you for the careful review of `setup.py`, and for offering to rebase or split the changes.

All three reported problems have now been resolved through the current packaging and compiler refactor: the Python-version message uses correct interpolation, compiler failures raise proper exception types, and the old bare exception handling has been removed. The macOS compiler logic itself has also moved into a tested `build_config.py` implementation.

A rebase is therefore no longer necessary. We are closing this PR as superseded, with thanks for identifying valid problems in the previous implementation.

PR #611, opened 2026-03-18
perf: split update_e and update_h into per-component kernels to reduce warp divergence

### LINKED ISSUE #608: warp divergence in update_e and update_h kernels in fields_updates_gpu.py
hey 
i was going through the kernels_template_fields.py I noticed that the update_e kernel handles all three field components (Ex, Ey, Ez) in a single kernel with separate boundary condition checks for each component. Threads near domain boundaries will take different branches depending on their (i,j,k) position, causing warp divergence
.similalry in update_e_dispersive_A, the inner pole loop iterates MAXPOLES times for all threads even if a material has fewer active poles .threads assigned to materials with fewer poles diverge more than them.
i think  launching another kernel  for each field component so boundary threads are isolated, or to pad the pole loop so all threads always iterate the same number of times.

SouravVerma-art:
Each combined kernel handled all 3 field components with different
boundary conditions, causing warp divergence for threads near domain
boundaries. Splitting into per-component kernels isolates boundary
threads so all threads in a warp take the same branch.

- update_e split into update_ex, update_ey, update_ez
- update_h split into update_hx, update_hy, update_hz
- Each kernel only receives the field arrays it needs
- model_build_run.py updated to get and call the new kernel functions

Closes #608

# PR Description

##  Related Issue  #608

Closes #

## Type of change

## Checklist

### DIFF
--- a/gprMax/fields_updates_gpu.py
+++ b/gprMax/fields_updates_gpu.py
@@ -37,13 +37,13 @@ __device__ __constant__ $REAL updatecoeffsH[$N_updatecoeffsH];
 // Electric field updates - standard materials //
 /////////////////////////////////////////////////
 
-__global__ void update_e(int NX, int NY, int NZ, const unsigned int* __restrict__ ID, $REAL *Ex, $REAL *Ey, $REAL *Ez, const $REAL* __restrict__ Hx, const $REAL* __restrict__ Hy, const $REAL* __restrict__ Hz) {
+__global__ void update_ex(int NX, int NY, int NZ, const unsigned int* __restrict__ ID, $REAL *Ex, const $REAL* __restrict__ Hy, const $REAL* __restrict__ Hz) {
 
-    //  This function updates electric field values.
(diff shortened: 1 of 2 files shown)


Codebruh-sudo:
hey , im also a GSOC aspirant . i saw the code and you have done a great work  . Splitting the kernel in component is a smart way to eliminate the boundary condition divergence. but i have some questions or concern . I thick there are some compilation errors . and yes i also noticed there is missing index calculation in update_ex and update_hx. 
 and also can you provide the performance benchmarks of the code  before and after check Warp Divergence Report also check register pressure etc  . i think you should check the code again before merging . it would be helpful. you can tell me if Iam wrong .

Codebruh-sudo:
hey , @SouravVerma-art  update_ex and update_hx are missing the i, j, k and i_ID index computations that were present in the original update_e/update_h kernels these variables are used in .the if condition and field accesses but never declared, which will cause a compile error. Also update_ez there is no closing braces.

PR #614, opened 2026-03-19
Reduce redundant integer multiplication in PML kernel index calculations in pml_updates_electric_HORIPML_gpu.py.

Codebruh-sudo:
hey ,
 i have raised this pr because of the repeated NX_PHI * NY_PHI * NZ_PHI and NY_PHI * NZ_PHI products in local variables across all PML electric field update kernels. Currently these integer multiplications are recomputed 4 and 3 times respectively per thread per kernel launch. The fix introduces two local variables per PHI block . volume and plane. that are reused throughout the index calculation. i hope this fix is good . if iam wrong with this you can tell me .I am always ready to improve .

### DIFF
--- a/gprMax/pml_updates/pml_updates_electric_HORIPML_gpu.py
+++ b/gprMax/pml_updates/pml_updates_electric_HORIPML_gpu.py
@@ -47,17 +47,20 @@ __global__ void order1_xminus(int xs, int xf, int ys, int yf, int zs, int zf, in
     int idx = blockIdx.x * blockDim.x + threadIdx.x;
 
     // Convert the linear index to subscripts for PML PHI1 (4D) arrays
-    int p1 = idx / (NX_PHI1 * NY_PHI1 * NZ_PHI1);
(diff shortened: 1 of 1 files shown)


agianno:
Thanks for investigating the PML index calculations and preparing this optimization.

Since this PR was opened, the v4 `devel` branch has reorganized the accelerator kernels into shared CUDA/OpenCL/Metal templates. The current implementation already uses intermediate `rem1`/`rem2` and `jk1`/`jk2` values, avoiding the deeply repeated remainder calculations present in the legacy implementation modified here.

The remaining dimension products are compiler common-subexpression candidates, and no compiler-output analysis or representative benchmark has demonstrated a measurable improvement from storing them explicitly.

This PR targets `master` and modifies a legacy PML file that is no longer used by the v4 `devel` architecture. Porting it would require a new implementation across the shared accelerator templates, supported by profiling and physics-parity testing.

Given the v4 release scope and our decision to close the related speculative PML performance issues, we are closing this PR without merging. Thank you again for the contribution.

PR #617, opened 2026-03-20
Use reciprocal multiplication instead of division in PML kernels.

Codebruh-sudo:
hey , 
iam a GSOC 2026 applicant i was going through the codes
 there i noticed .the finite difference calculations divide by the spatial discretisation d on every thread .
dHz = (Hz[  ] - Hz[  ]) / dx; 
Since d is uniform and identical for all threads, this division can be replaced with a reciprocal multiplication computed once per thread.
$REAL i.nv_d = 1.0f / d;
dHz = (Hz[  ] - Hz[  ]) * inv_d;
Floating point division is significantly slower than multiplication on GPU.typically 4x slower on NVIDIA hardware. i have applied this change to all 12 kernels across xminus, xplus, yminus, yplus, zminus, zplus faces at both order1 and order2. 
I have applied changes in  these 4 files
- kernels_template_pml_electric_HORIPML_gpu.py
- kernels_template_pml_electric_MRIPML_gpu.py
- kernels_template_pml_magnetic_MRIPML_gpu.py
- kernels_template_pml_magnetic_HORIMPL_gpu.py
  if i am wrong you can suggest me . i am always ready to improve .

### DIFF
--- a/gprMax/pml_updates/pml_updates_electric_HORIPML_gpu.py
+++ b/gprMax/pml_updates/pml_updates_electric_HORIPML_gpu.py
@@ -59,7 +59,7 @@ __global__ void order1_xminus(int xs, int xf, int ys, int yf, int zs, int zf, in
     int k2 = ((idx % (NX_PHI2 * NY_PHI2 * NZ_PHI2)) % (NY_PHI2 * NZ_PHI2)) % NZ_PHI2;
 
     $REAL RA01, RB0, RE0, RF0, dHy, dHz;
-    $REAL dx = d;
+    $REAL inv_d = 1.0f / d ;
     int ii, jj, kk, materialEy, materialEz;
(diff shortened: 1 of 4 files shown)


agianno:
Thanks for identifying the repeated spatial-discretisation divisions and implementing this optimization.

The intended change is already present in the current v4 `devel` implementation. All four shared PML accelerator templates—electric and magnetic HORIPML and MRIPML—calculate the reciprocal once (`dx_inv`, `dy_inv`, or `dz_inv`) and multiply the field differences by it.

The v4 implementation is shared by CUDA, OpenCL, and Metal, whereas this PR targets `master` and modifies the legacy backend-specific PML files. It is therefore superseded by the current `devel` implementation and is not applicable to the release branch.

Closing this PR without merging. Thank you for the contribution.

PR #624, opened 2026-03-22
Improve CLI error handling for missing input file

SIRILEKKALA:
This PR improves the CLI user experience by adding validation for the input file path.

Previously, providing a non-existent input file resulted in a Python traceback, which can be confusing for users.

This change introduces an explicit check in the CLI to ensure the input file exists before execution. If the file is missing, a clear and user-friendly error message is displayed.

This prevents unnecessary crashes and improves usability, especially for new users.

### DIFF
--- a/gprMax/gprMax.py
+++ b/gprMax/gprMax.py
@@ -17,6 +17,7 @@
 # You should have received a copy of the GNU General Public License
 # along with gprMax.  If not, see <http://www.gnu.org/licenses/>.
 import argparse
+import os
 
 import gprMax.config as config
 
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for identifying the missing-input-file error handling problem.

The fix has now been reimplemented against current devel in #818, with regression tests confirming that a missing input produces a concise argparse error without a Python traceback. Your contribution is credited in the replacement commit.

As the original PR conflicts with current devel and has been superseded by #818, we are closing it. Thank you for identifying and addressing the issue.

PR #625, opened 2026-03-22
Fix negative n validation

SIRILEKKALA:
## Summary

This PR adds validation for the `-n` CLI argument to ensure that the number of simulation runs is a positive integer.

## Motivation

Currently, gprMax allows non-positive values (e.g., 0 or negative numbers) for the `-n` argument. This can lead to invalid or undefined simulation behavior, as the number of runs should logically be a positive integer.

## Changes

- Added validation in the CLI (`gprMax.py`) after argument parsing
- If `-n <= 0`, a clear and user-friendly error message is displayed using `parser.error`

## Example

Before:
gprMax model.in -n -5
→ Accepted, but leads to incorrect or undefined behavior

After:
gprMax: error: Number of runs (-n) must be a positive integer. Got: -5

## Impact

- Improves robustness and correctness of CLI input handling
- Prevents invalid simulation configurations
- Provides clearer feedback to users

## Type of change

## Checklist

### DIFF
--- a/gprMax/gprMax.py
+++ b/gprMax/gprMax.py
@@ -302,6 +302,9 @@ def cli():
     )
     args = parser.parse_args()
 
+    if args.n <= 0:
+        parser.error(f"Number of runs (-n) must be a positive integer. Got: {args.n}")
+
     results = run_main(args)
 
     return results
--- a/toolboxes/Plotting/plot_Ascan.py
+++ b/toolboxes/Plotting/plot_Ascan.py
@@ -21,6 +21,7 @@ import logging
 from pathlib import Path
 
 import h5py
+import matplotlib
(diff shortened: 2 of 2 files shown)


agianno:
Thank you for identifying the need to reject non-positive model counts.

Current devel now validates both `n` and the optional starting model number `i` in `SimulationConfig`. This covers both the command-line interface and the Python API, and regression tests verify zero, negative, and valid values. The plotting change bundled into this PR has also been superseded by the v4 toolbox modernisation.

The original fix is therefore no longer needed and the PR now conflicts with current devel. We are closing it as superseded in preparation for the v4 release. Thank you for reporting and addressing the original behaviour.

PR #626, opened 2026-03-22
Add --dry-run flag for validating input without running simulation

SIRILEKKALA:
## Summary

This PR introduces a new `--dry-run` CLI flag that allows users to validate input files without executing a full simulation.

## Motivation

Currently, users must run the full simulation to verify whether an input file is valid. This can be time-consuming and inefficient, especially when iterating on model configurations.

This feature enables a lightweight validation step, which is particularly useful for interactive and exploratory workflows.

## Changes

- Added a `--dry-run` flag to the CLI
- When enabled, the program validates input arguments and exits before running the simulation
- Provides a clear success message upon validation

## Example
gprMax model.in --dry-run
Output:
Dry run successful: input validated.

## Impact

- Improves usability and workflow efficiency
- Enables faster iteration for users working with simulation inputs
- Lays groundwork for interactive interfaces and reactive workflows

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/gprMax/gprMax.py
+++ b/gprMax/gprMax.py
@@ -300,9 +300,17 @@ def cli():
         default=args_defaults["log_all_ranks"],
         help=help_msg["log_all_ranks"],
     )
+    parser.add_argument(
+        "--dry-run", action="store_true", help="Validate input file without running simulation"
+    )
+
     args = parser.parse_args()
 
     results = run_main(args)
+
+    if args.dry_run:
+        print("Dry run successful: input validated.")
+        return
 
     return results
 


agianno:
Thank you for proposing a validation-only mode.

Unfortunately, this implementation does not provide a dry run: it calls `run_main(args)` before checking `args.dry_run`, meaning the complete model and simulation have already run when the success message is printed.

A genuine validation mode needs a defined stopping point in the model lifecycle—for example, after parsing and semantic validation, or after complete geometry construction—and needs tests demonstrating that the time-stepping solver is never entered.

Because this PR is functionally incorrect, conflicted, and not required for the v4 release, we are closing it. The feature can be reconsidered later as a separately designed capability.

PR #632, opened 2026-03-24
Fix macOS build: use Clang + libomp instead of GCC

DtDOW:
## Description

This PR updates the macOS build configuration to use Apple Clang with `libomp` for OpenMP support, instead of requiring Homebrew GCC.

On macOS, Clang is the default compiler but does not support OpenMP directly. This change adds support for `libomp`, which enables OpenMP with Clang. If `libomp` is not available, the build falls back to GCC as before.

## Changes

* Use Clang with `-Xpreprocessor -fopenmp` and link against `-lomp`
* Detect Homebrew prefix for Intel and Apple Silicon
* Keep GCC as a fallback option

## Testing

Ran:

```bash
python setup.py build
```

Build completed successfully on macOS.

## Notes

`libomp` can be installed with:

```bash
brew install libomp
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -182,7 +182,7 @@ else:
         libraries = []
 
     elif sys.platform == "darwin":
-        # Check for Intel or Apple M series CPU
+        # Detect Apple Silicon vs Intel for Homebrew prefix
         cpuID = (
             subprocess.check_output(
                 "sysctl -n machdep.cpu.brand_string",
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for proposing Clang and libomp support for macOS.

This requirement has now been addressed more comprehensively on the current devel branch. Compiler handling has moved into `build_config.py`, with support for explicit and environment-provided libomp prefixes, the active Python environment, Homebrew discovery, runtime linking, portable wheel builds, and dedicated tests. Both Intel and Apple Silicon macOS wheels are also covered by CI.

The original PR is now conflicted and has been superseded by that implementation. We are therefore closing it in preparation for the v4 release. Thank you for identifying and working on the macOS build issue and for providing evidence that this approach worked on Mac OS.

PR #633, opened 2026-03-25
Update setup.py for Python version and error handling

Codebruh-sudo:
In this PR i have fixed five bugs in setup.py that affect macOS builds with modern gcc, Python version validation, and error handling.    

1.`raise` with a string literal does nothing in Python 3 (macOS only)
  `raise('some message')` is valid Python syntax but does NOT raise an exception 
  it simply evaluates the string and discards it. Replaced with `raise RuntimeError(...)`.

  2. rpath extraction breaks for gcc 10+ (macOS only)
  `gccpath[-1].split(os.sep)[-1][-1]` extracts only the last *character* of the
  gcc binary name. For `gcc-10` this gives '0' instead of '10', producing an
  invalid rpath and causing a linker failure.
  Fixed by using `.split('-')[-1]` to extract the full version string.

  3. Glob pattern gcc-[10-11]does not match versions 10/11 as intended(macOS only)
  `[10-11]` is a character class matching single chars `0` or `1`, not a version
  range. This also misses gcc-12, gcc-13, gcc-14 which are common on modern
  Homebrew. Fixed with gcc-1[0-9]* to correctly match all two-digit gcc versions.

  4. Python version check inconsistent with `python_requires`
  The early exit checked for < (3, 4) while `python_requires` was set to ">3.6".
  Updated the check to < (3, 6) and corrected ">3.6" to ">=3.6" so Python
  3.6 is not incorrectly excluded.

  5. distutils fallback fails on Python 3.12+
   distutils was removed in Python 3.12. The fallback  from distutils.core import
  setup` will always raise an ImportError on modern Python. Replaced with a clear
  error message directing users to install  setuptools.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,8 +19,7 @@
 try:
     from setuptools import setup, Extension
 except ImportError:
-    from distutils.core import setup
-    from distutils.extension import Extension
+    raise ImportError('gprMax requires setuptools. Install it with: pip install setuptools')
 
 try:
     import numpy as np
@@ -52,8 +51,8 @@ with open('README.rst', 'r', encoding='utf-8') as fd:
     long_description = fd.read()
 
 # Python version
-if sys.version_info[:2] < (3, 4):
-    sys.exit('\nExited: Requires Python 3.4 or newer!\n')
+if sys.version_info[:2] < (3, 6):
+    sys.exit('\nExited: Requires Python 3.6 or newer!\n')
 
 # Process 'build' command line argument
 if 'build' in sys.argv:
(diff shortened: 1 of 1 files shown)


DtDOW:
you can also refer to PR #632 where I tried upgrading from only GCC dependant to clang too as M series generally use clang.

Codebruh-sudo:
hey , @DtDOW Hi @DtDOW, thanks for pointing out PR #632 The Clang + libomp approach is a great idea.                                                                                                                                                                                                                  As you can see here  my PR targets master and focuses on fixing the existing GCC detection logic for those users who already have GCC installed.
 Your approach in #632  it removes the GCC dependency, which is the cleaner and a long term solution for  macos.
 It might be worth the maintainers considering merging both 
 my fixes as a short-term patch for GCC users, and your Clang+libomp path as the preferred approach going forward on macOS. Alternatively, the two could be combined into a single PR that tries Clang first and falls back to GCC. you can tell me if am wrong.

DtDOW:
@Codebruh-sudo agreed, clang can be default and GCC can be backup. It's upto maintainers what they prefer now.

agianno:
Thank you for identifying these setup issues. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. The PR now conflicts with current master, and resolving it as submitted would overwrite newer packaging changes, including dependency metadata. Since it is no longer a focused, easily mergeable v3 fix and v4 uses a different packaging system, we are closing it.

PR #641, opened 2026-03-29
feat: stabilize devel benchmarks and update deprecated API usage

dipayansardar73-decode:
Focus on maintaining the stability of the `devel` branch for high-performance benchmarking. Technical fixes to host information libraries and migrating legacy input file syntax to current standards.


DtDOW:
Hi Dipayan, great work overall, while going through the code I noticed a couple of thing : 

1)  In your 'tools/ml_runtime_predictor.py' whenever I am calling train_model() will the whole model will get trained again, that too on random data if no CSV file is provided thus maybe giving different predictions based on the data, also wont it be computationally costly ? I guess training once and then exporting a joblib file to predict would be more than enough? Or maybe it is by design and I am missing something ? 

2) In your readme you mention tools/visualize_3d.py to visualise 3D plots but it seems like it's not in tools section, maybe you forgot to upload it or do you plan to bring it in future word ?

dipayansardar73-decode:
Hi @DtDOW, I have reviewed the points you mentioned.

1. You are correct that retraining the model on every call is not the best approach. I am updating the predictor to use joblib for model persistence so it only trains when necessary.
2. I apologize for the missing file. I am adding the 3D visualization tool to the tools directory now as it was intended to be part of the utility suite.

DtDOW:
Hi Dipayan, once again good work, toh I would like to mention a couple of things 

1) You mention this eq ' df["runtime_s"] += df["grid_size"] * 0.1 + df["water_content"] * 50’ since this equation is linearly dependent on features, LinearRegression should perform better than RandomForestRegressor.

2) If you know the output formula (  ' df["runtime_s"] += df["grid_size"] * 0.1 + df["water_content"] * 50’  ), wont it be easier and more beneficial to directly predict result instead of training a model ? for eg

`def predict_runtime(grid_size, water_content):
    return random(10, 300)+ grid_size * 0.1 + water_content * 50 `
    
By creating an output using the input and then predicting the output using those input by ML algorithm you are fundamentally leaking your output data. Even if it scores well here in a situation where results are random and not governed by the given equation your model will fail.

dipayansardar73-decode:
Hello @DtDOW, I understand your points regarding the linear relationship and data leakage in the current implementation. I will replace the RandomForestRegressor with a LinearRegression model as it is better suited for this linear dependency. I also acknowledge that using a fixed formula for synthetic data is leading to data leakage, so I will modify the data generation to include more noise and varied sources to ensure the model does not simply learn the underlying formula. I will share the updated results with you as soon as the changes are complete.

agianno:
Thank you for the considerable amount of work represented by this PR.

After review, we cannot merge it in its present form. It combines 65 files covering legacy benchmark data, CI, release automation, plotting, 3D visualisation, water-content analysis, and an ML runtime predictor. Several of these areas are unrelated and would need independent design and review.

There are also concrete integration problems: the proposed workflows rely on build assumptions that do not apply to current devel, one CI command deliberately hides pytest failures, and the ML predictor is trained from synthetic outputs derived from its input features rather than measured gprMax performance data. The useful headless plotting work has meanwhile been superseded by the v4 toolbox modernisation.

As part of v4 release preparation, we are closing this broad and now-conflicted PR. Any still-useful idea should be resubmitted separately against current devel with measured data, focused tests, and one clearly defined responsibility per PR.

PR #649, opened 2026-03-31
Fix memory leak in multiscene calculations (Issue #389)

### LINKED ISSUE #389: Running out of memory
When calculating my multiscene configuration ( gprMax.run(scenes=scenes, n=len(scenes) ... ) Each my scene uses about 2gb of memory. It turned out that each scene's memory is added to the previous scene, and after a certain number of scenes, all of my computer's RAM (32gb) ran out and the calculation crashed. Is it possible to provide a mechanism for clearing memory during multiscene calculation? I read that similar problems were in the release version of gprMax when using GPU, but I'm using the new python API and not using GPU.

One more question : When using multiscene, output files are created with numbers 1, 2, etc., how do I pass the parameters of a specific scene to write to the output file name or to the output file attributes of a specific scene?

kallal79:
- Add import gc for garbage collection
- Enhanced memory management to prevent accumulation between models
- Always cleanup when geometry_fixed=False to prevent memory leaks
- Force garbage collection after each model deletion
- Maintains backward compatibility with geometry_fixed optimization

Fixes: https://github.com/gprMax/gprMax/issues/389
 

## Checklist

### DIFF
--- a/gprMax/model_build_run.py
+++ b/gprMax/model_build_run.py
@@ -18,6 +18,7 @@
 
 import datetime
 from importlib import import_module
+import gc
 import itertools
 import os
 import psutil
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for investigating the multiscene memory issue. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. Current master already deletes the grid in the relevant path, while this PR adds explicit garbage collection without a regression test demonstrating which retained reference it resolves. We cannot safely treat this as a verified fix, and the model lifecycle has changed in v4, so we are closing the PR.

PR #653, opened 2026-03-31
Add initial pytest unit tests for gprMax.utilities (atoi + natural_keys)

TanyaC07:
This adds basic unit tests for two simple utility functions:
- atoi()
- natural_keys()

These functions are used for natural/human sorting of strings (e.g. filenames with numbers).

Part of efforts toward building a comprehensive test suite for gprMax.

### DIFF
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -1,0 +1,25 @@
+import pytest
+from gprMax.utilities.utilities import atoi, natural_keys
+
+def test_atoi():
+    """Test atoi function for string to int conversion."""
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for adding these tests. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. These tests target the legacy utilities rather than correcting a release-critical problem, and the test structure has changed in v4. We are therefore closing this PR. Any missing v4 coverage should be proposed against devel or the new gprMax version once it is released.

Issue #600, opened 2026-03-15
Implement `pre-commit` configuration and automated linting workflow


subhamkumarr:
### Description
The current PR template asks contributors:
`- [ ] Did pre-commit passed all checks?`

However, there is no `.pre-commit-config.yaml` file currently present in the repository to actually define or run these checks locally, nor is there CI automation enforcing them. 

### Proposed Changes
To modernize the developer workflow and enforce baseline code quality, I propose adding a standardized `.pre-commit-config.yaml` file to the repository root.

To avoid disrupting `git blame` history with massive reformatting (`black`/`flake8`) right now, this initial configuration will stick to safe, structural checks:
- `trailing-whitespace`
- `end-of-file-fixer`
- `check-yaml`
- `check-merge-conflict`

I will also add a lightweight GitHub Actions workflow (`.github/workflows/lint.yml`) that automatically runs `pre-commit run --all-files` on PRs to prevent malformed files from being merged. I am happy to open a PR for this!

amitraghvan:
I'm addressing this as part of my GSoC 2026 proposal for 
Project 4. I've drafted a .pre-commit-config.yaml with flake8, 
isort, and standard hooks — specifically configured for gprMax's 
codebase (E402 ignored for conditional pycuda/mpi4py imports, 
200-char line limit for GPU kernel files). PR coming soon.

agianno:
file exist

Issue #598, opened 2026-03-15
Implement Basic GitHub Actions CI Pipeline for Automated Testing


subhamkumarr:
### Description
Currently, `gprMax` does not have an automated Continuous Integration (CI) pipeline set up via GitHub Actions. Maintainers and contributors must rely on local suite execution to verify code changes.

To modernize the development workflow and ensure cross-platform stability, I am proposing the addition of a foundational [.github/workflows/ci.yml](cci:7://file:///c:/Users/Subham/OneDrive/Desktop/gprMax/gprMax/.github/workflows/ci.yml:0:0-0:0) file.

### Proposed Changes
Implement a GitHub Actions matrix workflow that automatically:
- Triggers on PRs and pushes to `master`.
- Tests against `ubuntu-latest`, `macos-latest`, and `windows-latest`.
- Installs necessary system dependencies (e.g., `libomp` for macOS).
- Builds the Cython extensions in-place (`python setup.py build_ext --inplace`).
- Executes the `pytest` suite.

This will significantly reduce the manual review burden by automatically catching regressions across different operating systems before code is merged. I am happy to open a PR for this!

amitraghvan:
Hi @subhamkumarr, thanks for opening this issue! I've been studying the gprMax codebase 
for my GSoC 2026 proposal (Project 4: CI/CD Automation) and I've identified a few 
specific challenges for implementing this CI pipeline that I'd like to share:

1. macOS compiler detection:
The current `setup.py` (lines 120-141) uses `glob('/opt/homebrew/bin/gcc-[4-9]*')` to 
find gcc on Apple Silicon and `/usr/local/bin/gcc-*` for Intel Macs. On GitHub Actions 
macOS runners, we'll need `brew install gcc` first since only clang is pre-installed 
(and clang doesn't support OpenMP).
2. Headless test execution:
`tests/test_models.py` uses `matplotlib.pyplot` which requires a display backend. 
In CI (headless), we'll need to force `matplotlib.use('Agg')` — this is also 
related to issue #621.

3. Conda vs pip:
Since gprMax uses `conda_env.yml` for dependency management, the CI should use 
`conda-incubator/setup-miniconda` action rather than plain `pip install`.

I'm planning to work on this as part of my GSoC proposal and would love to 
collaborate. Happy to open a draft PR with a proof-of-concept workflow if that 
would be helpful!

agianno:
[...]

PR #655, opened 2026-04-01
build: add missing setup.cfg with flake8 and pytest configuration

amitraghvan:
## Problem

The project README explicitly references `setup.cfg` as the location for flake8 configuration, but this file does not exist in the repository. This causes two concrete issues for contributors:

- Running `flake8` without a config file causes it to fall back to defaults, which do not match gprMax's codebase conventions (200-character line length used in GPU kernel files).
- - Running `pytest` emits `PytestUnknownMarkWarning` for every test marked with `@pytest.mark.slow`, `@pytest.mark.gpu`, or `@pytest.mark.mpi`, because these markers are not registered anywhere.
## Solution

Add the missing `setup.cfg` with two sections:

- `[flake8]` - max-line-length 200, E402 ignored for conditional pycuda/mpi4py imports, W503 ignored as style preference.
- - `[tool:pytest]` - Registers the slow, gpu, and mpi markers, eliminating PytestUnknownMarkWarning.
## Notes

This PR targets the `devel` branch as per project contribution guidelines. No functional source changes - purely additive configuration.

Supports #600 and #598.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,16 @@
+[flake8]
+max-line-length = 200
+ignore = E402, W503
+exclude =
+    .git,
+    __pycache__,
+    build,
+    dist
+
+[tool:pytest]
+testpaths = tests
+addopts = -v --tb=short
+markers =
+    slow: marks tests as slow (full simulation runs - deselect with '-m "not slow"')
+    gpu: marks tests that require a CUDA-capable GPU
+    mpi: marks tests that require an MPI installation


agianno:
Thank you for investigating the missing test and lint configuration.

The current devel branch now registers pytest markers and other tool configuration in `pyproject.toml`, which is the configuration source used for v4. The old README reference to `setup.cfg` has also been removed. Adding a separate `setup.cfg` would now duplicate configuration and could allow the two files to diverge.

The underlying need has therefore been superseded, so we are closing this PR in preparation for the v4 release. Thank you for the contribution.

PR #656, opened 2026-04-01
tests: add conftest.py with headless matplotlib backend fix

### LINKED ISSUE #621: plot_Ascan does not handle headless environments properly
When running plot_Ascan in headless environments (e.g., WSL, remote servers),
the plotting fails due to unavailable GUI backends (Qt/Wayland).

Example:
python -m toolboxes.Plotting.plot_Ascan examples/cylinder_Ascan_2D.h5

Produces warnings like:
"FigureCanvasAgg is non-interactive, and thus cannot be shown"

In such cases:
- plt.show() is still called
- No plot is displayed
- Output is not clearly saved

Proposed solution:
- Detect headless environments using matplotlib backend
- Avoid calling plt.show() when GUI is unavailable
- Automatically save plots when running headless
- Provide clear feedback message

This would improve usability for users working in WSL, servers, and CI environments.

amitraghvan:
## Problem

Running `pytest` on the current gprMax test suite fails in any headless environment because `test_models.py` imports `matplotlib.pyplot` at module level, which immediately attempts to connect to a display server:

```
_tkinter.TclError: no display name and no DISPLAY variable
```

This crash prevents the entire test suite from being collected by pytest - meaning zero tests run in CI.

## Solution

Add `tests/conftest.py` which pytest loads automatically before collecting any tests:

1. Forces the `Agg` matplotlib backend before any import of pyplot - works in all headless environments.
2. 2. Provides shared path fixtures for test model directories (models_basic_path, models_advanced_path).
## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)
## Checklist

* [x] I have performed a self-review of my code.
* [ ] * [x] My changes generate no new warnings.
* [ ] * [x] The title of my pull request is a short description of my changes.
Closes #621

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,37 @@
+"""
+Pytest configuration and shared fixtures for the gprMax test suite.
+
+This file is automatically loaded by pytest before any tests are collected.
+It performs two tasks:
+
+1. Forces matplotlib to use the non-interactive 'Agg' backend *before* any
+   test module imports pyplot. Without this, tests crash in headless CI
+   environments (Linux runners, Docker containers) with:
+       "cannot connect to X server" or "_tkinter.TclError"
+   Fixes #621.
+
+2. Provides shared path fixtures so individual test modules do not need to
+   hard-code paths to the test model directories.
+"""
+import os
+
+import matplotlib
+matplotlib.use('Agg')  # noqa: E402 — must be called before pyplot is imported
+
+import pytest  # noqa: E402
+
+
+# ---------------------------------------------------------------------------
+# Path fixtures
+# ---------------------------------------------------------------------------
+
+@pytest.fixture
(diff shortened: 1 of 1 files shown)


agianno:
Thanks for looking at this. Forcing the Agg backend for the test suite is a reasonable safety net in general, but it doesn't actually address #621 as filed — that issue is about the toolboxes/Plotting/*.py CLI tools not handling a headless backend when run directly by a user, not about the pytest suite itself crashing (no file under tests/ currently imports matplotlib.pyplot, so there's nothing live for this to prevent there). It also adds models_basic_path/models_advanced_path fixtures pointing at tests/models_basic/tests/models_advanced, which don't exist — the real model directories are under testing/, not tests/. The actual #621 fix landed in #732. Closing this in favour of that.

Issue #591, opened 2026-03-11
Allow to run simulations on GPU even if vram is insufficient


carrrrrrrrot:
Hello. 

I've found out, by commenting the code that throws the out of memory error (line 240/241 in grid.py), that simulations can be run on the gpu even if vram is insufficient. Although this incurs in a peformance penalty, since some of the memory spills into system ram, to which the gpu has much slower access, it's still orders of magnitude faster than running on cpu.

The cpu I'm using is a xeon E5-1660v3, with 64GB of ram, the gpu is an rtx 3060, with 12GB of vram, and the input file I've used for testing requires 16gb of memory. Running the simulation on the cpu required more than 7 hours, while on the gpu it took less than 30 minutes. 

I imagine that with even larger simulations, which would require a larger portion of system ram to be used, this advantage may grow thinner. However this would still be extremely useful in intermediate scenarios. The error could be replaced with a warning that notifies the user about possible performance degradation, without stopping the execution.

Might be worth mentioning that on more modern systems, which support faster pcie standards and features such as resizable bar, the transfer between system ram and gpu is a lot faster, so the performance penalty should be significantly smaller. 

I understand that this may be more of a feature request rather than an issue, so if this report would have been better placed somewhere else I apologize.

DtDOW:
Hi, Great observation !

I created a PR for it with 3 different 'run modes': 

'strict'      - raise an error and stop (default, preserves original behaviour)
'overflow'    - warn and continue; model spills into system RAM (slower)
'fallback_cpu'- warn and fall back to CPU execution

in 'strict' settings it continues as it was before, in 'overflow' it spills into RAM and in 'fallback_cpu' it completely shifts to CPU

agianno:
Thank you for reporting this and for providing the useful performance measurements.

We reviewed the current v4 `devel` implementation. It no longer rejects a model solely because its estimated memory requirement exceeds the reported GPU capacity.

The current behaviour is:

- Models without snapshots do not undergo the old pre-emptive device-memory rejection.
- When snapshots are present, exceeding the device-memory estimate produces a warning rather than terminating the simulation.
- If the non-snapshot working set fits on the device, snapshot storage is transferred back to the host during the simulation.
- The backend and driver are allowed to attempt the allocation, so platforms that support oversubscription or unified memory can continue.
- Metal uses unified system memory and is governed by the host-memory capacity check.

[...]

PR #664, opened 2026-04-03
Run strategies for when VRAM is insufficient. 

DtDOW:
Build on #591 : 

In case GPU vram isn't sufficient for run it introduces 3 option to continue the run based on user preference :  

1)  'strict'  - raise an error and stop (default, preserves original behaviour)
2) 'overflow'    - warn and continue; model spills into system RAM (slower)
3) 'fallback_cpu'- warn and fall back to CPU execution 

User can decide based on their personal preference ( default is 'strict' mode  thus no change in default behaviour)

### DIFF
--- a/gprMax/grid.py
+++ b/gprMax/grid.py
@@ -105,6 +105,12 @@ class FDTDGrid(Grid):
         # GPU object
         self.gpu = None
 
+        # GPU memory strategy when VRAM is insufficient:
+        # 'strict'      - raise an error and stop (default, preserves original behaviour)
+        # 'overflow'    - warn and continue; model spills into system RAM (slower)
+        # 'fallback_cpu'- warn and fall back to CPU execution
+        self.gpu_memory_strategy = 'strict' 
+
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for developing these possible strategies for handling insufficient reported VRAM.

The v4 `devel` branch has since changed both the memory-checking architecture and the relevant behaviour. It no longer rejects a model solely because its estimated device-memory requirement exceeds the reported GPU capacity.

The current implementation permits the backend and driver to attempt the allocation, warns where appropriate, and transfers snapshot data to the host when the non-snapshot working set fits. Metal uses unified system memory and follows the host-memory capacity check.

This PR modifies the legacy `gprMax/grid.py` implementation on `master`, which no longer corresponds to the v4 grid, configuration, or solver architecture. Its new strategy setting is consequently not suitable for merging into `devel` as written.

Issue #591 has been closed as satisfied by the current v4 behaviour. Closing this PR as superseded.

PR #666, opened 2026-04-06
Security: Implement a sandbox for #python scripting blocks

sahilshingate01:
Hi gprMax team,

I noticed that the current implementation of #python blocks in .in files allows for arbitrary code execution on a user's system. Since researchers frequently share simulation models, this represents a potential security risk where a malicious file could perform unintended actions.

Even though these blocks are deprecated in the devel branch, I wanted to provide a "root cause" fix to ensure the community is protected until the feature is fully removed.

What this PR does:

Safety Scan: It now scans the Python code before it runs using the ast module. It proactively blocks any import statements or dangerous calls like open(), exec(), or eval().
Restricted Environment: I've moved the execution into a strict sandbox. By default, the script has zero access to standard Python built-ins. It only has access to a safe "whitelist" of functions (like range, math, and numpy) that are actually needed for GPR modeling.
Private Data Protection: It blocks any attempts to access private Python attributes (like __class__) to prevent sandbox breakouts.
This keeps the project's powerful scripting capabilities intact for current users while making it significantly safer to run shared models from the internet or colleagues.

Note: I've focused this PR on the devel branch as requested, specifically updating the new hash_cmds_file.py logic.

### DIFF
--- a/gprMax/hash_cmds_file.py
+++ b/gprMax/hash_cmds_file.py
@@ -17,7 +17,9 @@
 # You should have received a copy of the GNU General Public License
 # along with gprMax.  If not, see <http://www.gnu.org/licenses/>.
 
+import ast
 import logging
+import os
 import sys
 from io import StringIO
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for identifying the security implications of executing legacy `#python` blocks.

We cannot merge this implementation. It contains an immediate runtime error because `codeout` is read without first being assigned. It would also prevent variables created in one Python block from being propagated through the existing user namespace.

More fundamentally, an AST blacklist combined with selected Python built-ins is not a secure in-process Python sandbox. There are many indirect ways to reach functionality that a blacklist does not anticipate, so merging this could give users a false sense that untrusted model files are safe.

The legacy `#python` feature is deprecated. We will handle the remaining concern through explicit trust warnings and its eventual removal rather than claiming to sandbox arbitrary Python. We are therefore closing this PR, but thank you for drawing attention to the issue.

PR #667, opened 2026-04-06
Refactor: Removing global state reliance for thread-safe simulation runs

sahilshingate01:
Hello gprMax team,

I've been working on an architectural refactor for gprMax to address a key piece of "architectural debt": the engine's heavy reliance on global configuration state. As you're working towards a major release on the devel branch, I thought this would be a great time to implement this improvement.

What’s the problem?
Previously, many core parts of the simulation (like the grid, model, and source definitions) looked at global variables or "singletons" to find their settings (data types, solver type, etc.). This caused a major bottleneck:

No Parallelism: You couldn't safely run multiple simulations in separate threads within the same Python script because they would all try to fight over those same global settings.
Hard to Test: Testing small pieces of code was difficult because they were "wired" into the global state.
What have I changed?
I’ve implemented what’s known as "Dependency Injection." In simple terms, instead of a function reaching out to a global variable for settings, the settings are now "injected" or passed directly to the function or object that needs them.

Key highlights of the refactor across 29 files:

Model-bound Settings: Each simulation model now carries its own configuration. When you create a Model, it gets its own sim_config and model_config.
Sources & Utilities: All source objects (dipoles, voltage sources, etc.) and utility functions (like memory checks) now get their configuration directly from the grid they belong to.
Thread Safety: Because each simulation is now isolated with its own configuration, this refactor is a direct prerequisite for enabling high-performance multi-threaded runs in the future.
Why does this matter?
This change makes gprMax much more modular and robust. It opens the door for developers to run massive batches of simulations concurrently in a single process without worrying about global state corruption.

I’ve focused these changes on the devel branch and ensured that the core logic remains intact while significantly cleaning up the internal data flow.

I would love your feedback on these improvements!


agianno:
Thank you for working on this. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. This is a large architectural refactor that overlaps substantially with the restructuring already undertaken for v4 and is not suitable for merging into the legacy branch. We are therefore closing this PR.

PR #669, opened 2026-04-11
feat: add --force-gpu flag to allow GPU memory spill bypass (#591)

### LINKED ISSUE #591: Allow to run simulations on GPU even if vram is insufficient

dipayansardar73-decode:
Fixes #591

This PR adds a --force-gpu configuration flag to bypass the rigid totalmem exception during building and execution.

If a user demands more memory than the GPU natively possesses, gprMax previously threw a hard GeneralError crash. This PR intercepts that specific crash when the flag is enabled, replacing it with a RuntimeWarning. This allows the host driver to dynamically spill excess allocations into System RAM instead, vastly expanding the size of simulation an individual card can run (albeit with expected performance penalties).

The flag has been plumbed directly through the main api() endpoint and threaded down into the grid.memory_check() handler.

**Note:** As requested in #591.

### DIFF
--- a/gprMax/gprMax.py
+++ b/gprMax/gprMax.py
@@ -64,6 +64,7 @@ def main():
     parser.add_argument('--geometry-fixed', action='store_true', default=False, help='flag to not reprocess model geometry, e.g. for B-scans where the geometry is fixed')
     parser.add_argument('--write-processed', action='store_true', default=False, help='flag to write an input file after any Python code and include commands in the original input file have been processed')
     parser.add_argument('--opt-taguchi', action='store_true', default=False, help='flag to optimise parameters using the Taguchi optimisation method')
+    parser.add_argument('--force-gpu', action='store_true', default=False, help='bypass GPU VRAM checks and allow memory to spill to System RAM')
     args = parser.parse_args()
 
     run_main(args)
@@ -82,7 +83,8 @@ def api(
     geometry_only=False,
     geometry_fixed=False,
     write_processed=False,
-    opt_taguchi=False
+    opt_taguchi=False,
+    force_gpu=False
 ):
     """If installed as a module this is the entry point."""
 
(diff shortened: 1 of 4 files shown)


DtDOW:
Hi Dipayan ! I filed PR #664 which also addresses #591 . My implementation offers three configurable strategies : 

1) strict (error out, default)
2) overflow (spill to system RAM)
3) fallback_cpu (gracefully fall back to CPU).

This gives users finer control over the tradeoff between performance and execution continuity, including the CPU fallback case that this PR doesn't currently cover.

dipayansardar73-decode:
Ah nice catch! Your implementation with the three configurable strategies is definitely way more robust. I didn't spot PR #664 before throwing this up. I'll close this one out in favor of yours. Great work!

PR #670, opened 2026-04-11
test: wrap legacy scripts in main block to prevent execution on import (Resolves #627)

### LINKED ISSUE #627: Legacy test scripts execute at import time, which blocks pytest-friendly test collection
## Description

The legacy test scripts in `tests/` currently execute work at **import time** instead of only when run directly as scripts. 

Specifically:
- `tests/test_models.py` starts running model-comparison logic immediately when imported.
- `tests/test_experimental.py` parses CLI arguments immediately when imported.

This makes the files awkward to use with `pytest` and future CI automation, because test runners typically import modules first for **discovery/collection**.

## Why this matters

Even before a broader CI/testing pipeline is added, these files would be easier to automate if their executable logic were moved behind a `main()` function and guarded with:

```python
if __name__ == "__main__":
    main()

## Scope

This issue is not about adding full CI or converting the whole suite to pytest right away. It is only about making the existing legacy test scripts safer to import and easier to build future automation on top of.

dipayansardar73-decode:
Fixes #627

### Description
This PR resolves the issue where Pytest collection blocks aggressively because legacy test configurations ran massive simulations synchronously at import-time.

By sweeping the global script execution logic inside `tests/test_models.py` and `tests/test_experimental.py` under standard `def main():` scopes guarded by `if __name__ == "__main__":`, we completely sidestep the bug. Both scripts are now 100% safe to import by CI pipelines without exploding, while retaining full backwards compatibility for users who trigger the scripts manually via the command line.

### DIFF
--- a/gprMax/gprMax.py
+++ b/gprMax/gprMax.py
@@ -64,6 +64,7 @@ def main():
     parser.add_argument('--geometry-fixed', action='store_true', default=False, help='flag to not reprocess model geometry, e.g. for B-scans where the geometry is fixed')
     parser.add_argument('--write-processed', action='store_true', default=False, help='flag to write an input file after any Python code and include commands in the original input file have been processed')
     parser.add_argument('--opt-taguchi', action='store_true', default=False, help='flag to optimise parameters using the Taguchi optimisation method')
+    parser.add_argument('--force-gpu', action='store_true', default=False, help='bypass GPU VRAM checks and allow memory to spill to System RAM')
     args = parser.parse_args()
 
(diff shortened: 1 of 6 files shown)


agianno:
Thank you for the contribution. We are preparing the gprMax v4 release, after which the current v3 master branch will be archived. This branch combines several separate changes—including legacy-test handling, packaging and GPU options—rather than providing a focused critical fix. The relevant areas have also changed substantially in v4, so we are closing this PR rather than merging it into master.

Issue #382, opened 2023-09-05
pyopencl event timing on macOS


craig-warren:
https://github.com/inducer/pyopencl/discussions/543

pratiktech28:
can i pick solve this issue and assign me with regards

agianno:
Looked into the linked pyopencl discussion. The root cause is confirmed to be Apple's OpenCL ICD giving unreliable kernel-event profiling data on macOS (not a pyopencl or gprMax bug) - the discussion's own maintainer notes "the Apple ICDs don't have a stellar reputation" for this.

The suggested workaround there - enqueue markers before/after the timed region and diff their profiled timestamps, rather than profiling individual kernel events directly - is exactly what gprMax already does in `OpenCLUpdates.time_start()`/`calculate_solve_time()` (gprMax/updates/opencl_updates.py).

The discussion also notes that even this marker-based approach can still show zero timing on GPU with the Apple ICD in some cases - that residual behaviour is a driver-level limitation with no application-side fix available. Since this only affects the reported benchmark `solvetime` figure and not simulation correctness, I'm closing this as "as good as it gets" given the upstream constraint. Please reopen if you hit a case beyond what's described here.

PR #671, opened 2026-04-12
Fix: OpenCL solve time returns 0 on Apple ICD with discrete GPU ( #382 ) 

DtDOW:
Builds on : #382 

Description : 

calculate_solve_time() was returning 0 on macOS with discrete AMD/NVIDIA GPUs because Apple's OpenCL ICD returns 0 for marker profile timestamps on those devices. 

The fix is simple is if the profile timestamps come back as 0 or negative, fall back to `time.perf_counter()`. This works correctly because event_marker2.wait() already blocks the host until the device is done, so the wall-clock reading is accurate.

Changes:

time_start() — records a wall-clock timestamp alongside the OpenCL marker
calculate_solve_time() — uses wall-clock as fallback when profile timestamps are zero

### DIFF
--- a/gprMax/updates/opencl_updates.py
+++ b/gprMax/updates/opencl_updates.py
@@ -19,6 +19,7 @@
 
 import logging
 from importlib import import_module
+import time
 
 import numpy as np
 from jinja2 import Environment, PackageLoader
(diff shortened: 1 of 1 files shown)


agianno:
Thank you for identifying the residual zero-timing behaviour produced by Apple’s OpenCL ICD.

The fallback has now been reimplemented against current devel in #819. Valid device profiling remains authoritative, while zero or negative device timestamps fall back to monotonic wall-clock timing. Regression tests cover both paths, and your contribution is credited in the replacement commit.

This affects performance reporting only and does not change simulation physics. As the original PR conflicts with current devel and has been superseded by #819, we are closing it. Thank you for the contribution.

Issue #673, opened 2026-04-15
No clear way to determine memory used from PyOpenCL unlike PyCUDA.


DtDOW:
## Problem

Currently, device (GPU) memory usage reporting in gprMax is only available for the CUDA backend. The OpenCL backend does not provide any mechanism to report device memory usage.

This is primarily because OpenCL does not expose a direct API equivalent to CUDA’s `mem_get_info` for querying memory usage at runtime.

As a result, users running simulations with OpenCL (e.g., on Apple Silicon or non-NVIDIA GPUs) always see:

Memory used (estimated): host + unknown for device

which is misleading and does not reflect the actual device memory usage.

pratiktech28:
can you pick the issue and assign me with regards

DtDOW:
I dont have the power to assign an issue, toh you can refer to #674 for my solution, if you have a better approach we can discuss it and give a PR with better solution.

agianno:
Thanks for raising this and for investigating the available OpenCL mechanisms.

We reviewed this against the current v4 `devel` branch. The standard OpenCL device queries expose total global-memory capacity, but they do not provide a portable equivalent of CUDA's runtime free/used-memory query. Some vendors provide extensions, but these are not available consistently across the devices supported by the OpenCL backend.

We also reviewed PR #674. Summing known PyOpenCL buffer sizes would measure part of gprMax's logical allocation footprint rather than actual device-memory usage. The proposed implementation does not include PML arrays and several optional v4 allocations, and maintaining a manual list would be fragile as new features are added. It also cannot account for driver/context overhead, allocation padding, memory residency, or unified-memory behaviour.

The current output explicitly reports the OpenCL device contribution as unknown, which is accurate given the portable API available. A future allocation-tracking system could be considered as a separate feature if there is sufficient demand, but it is not required for the v4 release.

Closing this as not planned because there is no reliable cross-platform implementation through standard OpenCL.

PR #674, opened 2026-04-15
adds device (GPU) memory usage for OpenCL

DtDOW:
builds on #673 

## Problem

Currently, memory usage reporting is only available for the CUDA backend via `mem_get_info`.  OpenCL does not provide a direct API to query device memory usage at runtime. 

## Implementation

Memory usage is estimated by summing the sizes of all allocated device buffers using `.nbytes`. The following categories are included:

1)  Core field arrays (E and H fields, ID grid)
2)  Dispersive material arrays (if present)
3) Source arrays (Hertzian, magnetic, voltage)
4)  Receiver arrays
5)  Snapshot arrays

The solver loop was updated to invoke `calculate_memory_used` in a backend-agnostic way.

## Validation

1)  Memory usage was verified to scale proportionally with domain size (number of grid cells).
2)  Increasing the domain dimensions resulted in expected increases in memory usage (~O(N) with respect to number of cells).

## Images 

Case 1 : 
Domain 2, 2 , 0.1 

Screenshot 2026-04-15 at 22 22 29

Screenshot 2026-04-15 at 22 22 42

Case 2 : 
Domain 20, 20, 0.1 

Screenshot 2026-04-15 at 22 23 40

Screenshot 2026-04-15 at 22 23 36

Increase by 100x ( 2x10, 2x10, 0.1x1 = 20, 20, 0.1 ) 
increase in memory : ~100x

### DIFF
--- a/gprMax/model.py
+++ b/gprMax/model.py
@@ -541,17 +541,15 @@ class Model:
 
         # Run solver
         solver.solve(iterator)
-
         # Write output data, i.e. field data for receivers and snapshots to file(s)
         self.write_output_data()
-
         # Print information about memory usage and solving time for a model
         # Add a string on device (GPU) memory usage if applicable
         mem_str = ""
         if config.sim_config.general["solver"] == "cuda":
             mem_str = f" host + ~{humanize.naturalsize(solver.memused)} device"
         elif config.sim_config.general["solver"] == "opencl":
-            mem_str = f" host + unknown for device"
+            mem_str = f" host + ~{humanize.naturalsize(solver.memused)} device (estimated, excludes driver overhead)"
 
         logger.info(
             f"Memory used (estimated): "
--- a/gprMax/solvers.py
+++ b/gprMax/solvers.py
@@ -94,7 +94,7 @@ class Solver:
 
             if isinstance(self.updates, MPIUpdates):
                 self.updates.halo_swap_electric()
-            if isinstance(self.updates, CUDAUpdates):
+            if isinstance(self.updates, (CUDAUpdates, OpenCLUpdates)):
                 self.memused = self.updates.calculate_memory_used(iteration)
 
         self.updates.finalise()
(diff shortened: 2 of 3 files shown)


agianno:
Thanks for tackling this - OpenCL's lack of a runtime memory-query API is a real gap, and summing allocated buffer sizes is a sound approach given that constraint. Two things worth fixing before this is ready to merge:

1. **PML device arrays aren't counted.** The summation covers core field arrays, dispersive material arrays, source arrays, receiver arrays, and snapshot arrays, but PML arrays (`ERA_dev`, `ERB_dev`, `ERE_dev`, `ERF_dev`, `HRA_dev`, `HRB_dev`, `HRE_dev`, `HRF_dev`, `EPhi1_dev`, `EPhi2_dev`, `HPhi1_dev`, `HPhi2_dev` - allocated per-slab in `PML.htod_field_arrays()`, `gprMax/pml.py`) live on each PML object in `self.grid.pmls["slabs"]`, not on `self.grid`/`self` directly, so they're never summed here. For the default PML setup (10-cell thickness on all 6 domain faces), this can be a significant fraction of total device memory for small-to-medium models, not just "driver overhead" as the docstring's caveat suggests - worth iterating over `self.grid.pmls["slabs"]` and summing each slab's arrays too.

2. There's a leftover debug statement: `print(f'Total Memory : {total}')` in `calculate_memory_used()` - this would print on every OpenCL run and should be removed before merging.

Happy to help iterate on this if useful.

agianno:
Thanks for implementing and testing this approach.

After reviewing it against the current v4 `devel` branch, we have decided not to merge manual OpenCL buffer summation for the v4 release. The implementation does not currently count PML device arrays or several newer optional allocations, and maintaining a complete attribute list would be fragile as the solver evolves.

Even with complete enumeration, the result would be gprMax's logical buffer footprint rather than actual device-memory usage: it would not include driver/context overhead, allocation padding, residency decisions, or unified-memory behaviour. Standard OpenCL does not provide a portable CUDA-style query for that information.

For now, gprMax will continue to report that OpenCL device-memory usage is unavailable. A future implementation would preferably use central allocation tracking rather than a manually maintained list.

Thank you again for the contribution. Closing this PR without merging.

PR #711, opened 2026-07-27
Add simulation progress tracker (Component 2): GSoC 2026

alphaleporus:
## Summary

Adds `toolboxes/Marimo/progress_tracker.py`. Launches a gprMax simulation as a subprocess, parses its tqdm output, shows live progress.

- **Launch:** subprocess with `--show-progress-bars`, stdout and stderr merged into one pipe.
- **Parsing:** background thread reads output line by line. Regex only matches real tqdm lines (requires the trailing `[...it/s]` bracket, otherwise stray text like "Model 1/1" matches too).
- **Live updates:** `mo.ui.refresh` polls a plain dict the thread writes to, mirrors it into `mo.state`. No callback API exists in gprMax for this (checked, nothing under `gprMax/` matches `callback` or `hook`).
- **Progress bar:** plain HTML/CSS. `mo.ui.progress_bar` isn't in my installed marimo version.

One correction from earlier notes: tqdm output is on stdout, not stderr. Confirmed by redirecting both separately on a real run, stderr came back empty.

## How to test

1. `marimo run toolboxes/Marimo/progress_tracker.py`
2. Pick a `.in` file with enough iterations to see it move, `cylinder_Ascan_2D.in` finishes in under a second.
3. Run, confirm the bar updates and ends green with the right iteration count.
4. Stop mid-run, confirm the neutral callout and no leftover process (`ps aux | grep gprMax`).

## Part of

GSoC 2026 Component 2. Next: production `bscan_dashboard.py`.

### DIFF
--- a/toolboxes/Marimo/progress_tracker.py
+++ b/toolboxes/Marimo/progress_tracker.py
@@ -5,8 +5,6 @@ Progress comes from parsing gprMax's tqdm output on stdout (not stderr -
 checked). No callback API exists for this in gprMax itself.
 """
 
-from __future__ import annotations
-
 import marimo
 
 __generated_with = "0.23.8"
@@ -71,7 +69,7 @@ def _(mo):
                     "parsed from its tqdm output."
                 ),
                 input_browser,
-                mo.hstack([run_button, stop_button], gap="1rem"),
+                mo.hstack([run_button, stop_button], gap="1rem", justify="start"),
             ],
             gap="0.4rem",
         )
@@ -182,7 +180,10 @@ def _(get_progress, mo, refresh, run_lock, run_state, set_progress):
             "returncode": run_state["returncode"],
             "last_lines": list(run_state["last_lines"]),
         }
-    set_progress(snapshot)
+    # Only write when something changed. Setting identical state five times
+    # a second keeps the reactive graph busy for the life of the notebook.
+    if snapshot != get_progress():
+        set_progress(snapshot)
     p = get_progress()
 
     if not p["running"] and p["returncode"] is None:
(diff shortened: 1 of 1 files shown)


alphaleporus:
## Update, 21 August

Three fixes:

- The run and stop buttons were in an `mo.hstack` without `justify="start"`.
  marimo defaults to `space-between`, which pushed them to opposite edges of
  the container.
- The progress mirror called `set_progress` on every timer tick whether or not
  anything had changed, writing identical state five times a second for the
  life of the notebook. It now writes only on a real change.
- Removed an unused `from __future__ import annotations`, which sat between the
  docstring and the marimo import and tripped `marimo check --strict`.

`marimo check --strict` is now clean.

## Error handling, verified against a real failure

The proposal committed to handling a non-zero exit gracefully. That path was
exercised unintentionally while testing these fixes: a stale Cython build made
gprMax die on import. The notebook caught the non-zero return code, named it,
and surfaced the last lines of output including the real traceback, rather than
hanging or showing an empty progress bar. Screenshot below.

## Two findings about gprMax's progress output

This notebook merges stderr into stdout and parses that.

A bare `\d+/\d+` regex is not safe for parsing it. gprMax prints `Model 1/1,
input file: ...` and `Writing geometry view file 1/1,` as ordinary status
lines. The pattern here requires tqdm's trailing bracket, which separates the
progress bar from the prose.

There is no callback or hook API for progress in gprMax:
`grep -rn "callback|hook" gprMax/ --include="*.py"` returns nothing. Parsing the
output stream is the only option available, not a workaround for one.
Progress Tracker (1)
image

agianno:

PR #723, opened 2026-07-29
CUDA implementation of discrete plane wave (TF/SF) sources, including 2D TM/TE  - GSoC 2026

Codebruh-sudo:
##SUMMARY

This adds a CUDA implementation of the discrete plane wave sources, so DPW
models can now run on the GPU. It covers dispersive
and non-dispersive materials, and 2D TM/TE modes as well as 3D.

The 2D support follows the skip_axis approach used in applyTFSF*- in a 2D
model the invariant axis has no faces to inject through, so the corresponding
face pair is skipped in each of the CUDA TF/SF face launchers.

##THE TWO KERNELS MODULE

knl_planewave_updates.py -  the 1D auxiliary grid the DPW propagates
along. This is where the time is spent: one thread per cell, marching the
incident wave forward each timestep. It has standard (homogeneous) and axial
variants, each with dispersive and non-dispersive forms plus PML. The
dispersive kernels keep the polarisation state and coefficients as $COMPLEX
and use the GPRMAX_C* macros, so Debye, Lorentz and Drude plane-wave media
all work  the same pattern `knl_fields_updates.py` uses for the bulk updates.
The axial variants take the material coefficients as pointer arguments instead
of through constant memory, which keeps them clear of the 64 KB limit when a
model has many materials.

knl_tfsf_injection.py  -the face corrections that turn the 1D wave into
a 3D one. Each of the six faces of the TF/SF box needs its own kernel for the
electric and magnetic updates, in both standard and axial forms. They're small
and thread-light compared with the 1D updates, but they're where the total
and scattered field regions are stitched together, and where skip_axis does
its work in 2D.
On the host side, cuda_updates.py gains the kernel setup, device array
uploads and launch logic, and `cuda_grid.py` gains the DPW array uploads. The
axial kernels take the material coefficients as pointer arguments rather than
through constant memory, which keeps them clear of the 64 KB constant-memory
limit  htod_mat_coeff_arrays() uploads those.

solvers.py now calls the plane-wave updates for CUDA Updates as well as
CPU Updates, and the CUDA guard has been lifted from the three plane-wave
user objects in cmds_multiuse.py. OpenCL and Metal remain blocked, since
those backends aren't implemented yet.

##VALIDATION

CPU vs GPU comparison across all three plane-wave commands, in 3D, 2D TM and
2D TE, with no dispersion, with Debye, and with Lorentz and Drude  each both
as a scatterer and as the plane wave's own medium, single-pole and multipole. all the variants has been vaildated .
Receivers and snapshots both checked, in single and double precision. all the variants has been vaildated .

Most cases come out bit-identical,the rest differ only by floating-point
rounding, which I confirmed by rerunning in double precision and watching the
difference fall away with the mantissa.

For the 2D runs I also checked mode purity  the components that shouldn't
exist in each mode come out identically zero on both backends, across whole
snapshot volumes rather than just at receivers. That was the check that
mattered most, since a 2D model quietly running as 3D would still show CPU and
GPU agreeing with each other.

##PROFILING
The current CUDA kernel structure and thread execution bounds are well-optimized with minimal room for further tuning at this stage.


agianno:
Independently ran 24 CPU-vs-CUDA scenarios myself on real hardware (3D/2D-TM/2D-TE × none/Debye/Lorentz/Drude × single/double): 24/24 pass, agreeing to floating-point rounding levels (~1e-7 single, ~1e-10 double). Mode purity confirmed exact — dead components are literally zero on both backends, not just small.

Codebruh-sudo:
Thanks! and thanks for running the independent validation, that was reassuring to see. Happy to pick up anything that comes out of it, and let me know what would be most useful next.

PR #720, opened 2026-07-28
GSoC / Component 3: B-scan dashboard (live monitoring and load-files assembly)

alphaleporus:
**Builds on #696 (please merge that first).**

This branch is stacked on gsoc/component-3-ascan-dashboard, so the diff below also shows the 4 files from #696 (h5_reader.py, ascan_dashboard.py, test_h5_reader.py, requirements.txt). New in this PR: bscan_dashboard.py, trace_matrix.py, tests/test_trace_matrix.py.

## Summary

Adds `toolboxes/Marimo/bscan_dashboard.py` and `toolboxes/Marimo/trace_matrix.py`. Two ways to build a radargram from single-trace `.h5` files: watch one while gprMax is still writing it, or assemble one from files you already have.

- Live: `mo.ui.refresh` polls a directory on a selectable interval, reads only unseen `<basename><N>.h5` files each tick, appends to an accumulated matrix. Never rebuilds from scratch: checked by re-polling with no new files present and confirming zero traces get re-read.
- Load: any set of single-trace `.h5` files, ordered by source x-position read from each file's `srcs/src1` metadata. Falls back to file-selection order when position metadata isn't present.
- Core logic (`trace_matrix.py`) has no marimo dependency, same pattern as `h5_reader.py`. `process_trace()` validates and extracts one column, `stack_traces()` assembles an ordered list into a matrix. 15 tests in `tests/test_trace_matrix.py`, including a fixture set that specifically catches lexicographic-vs-numeric filename sorting (`1, 10, 11, 12, 2...` vs `1, 2, 3...12`) and a test that shuffles file-selection order and checks the assembled matrix is identical regardless.
- Both modes render Heatmap and 3D Surface (`go.Surface`) from the same matrix, plus receiver/component selection, a metadata banner (reuses `h5_reader.format_metadata_text`), time-window zoom, per-trace normalize (view only, CSV export stays unnormalized), and CSV/SVG/PDF export.

Found and fixed a real bug while building this: SVG/PDF export was computed eagerly on every render, which meant every poll tick during live monitoring re-ran two kaleido/Chrome invocations. 
`mo.download`'s `data` param takes a lazy callable. Fixed here, and the same bug turned out to be in `ascan_dashboard.py `too (its export code is where this file's was copied from), fixed separately in #696.

Bscan Dashboard (4)
Bscan Dashboard (3)

## How to test

1. `marimo run toolboxes/Marimo/bscan_dashboard.py`
[...]

PR #696, opened 2026-07-13
Add A-scan post-processing dashboard and h5_reader foundation (Compon…

alphaleporus:
## Summary
This PR replaces the static `plot_Ascan.py` script with our new reactive marimo dashboard. 

To keep the architecture clean, I split the work into two distinct layers: a framework-agnostic data reader and the marimo UI. 

### 1. The Backend (`h5_reader.py` & `test_h5_reader.py`)
I deliberately kept `marimo` imports completely out of the reader. It is pure Python and `h5py`. 
* Reads any number of v4 `.h5` files into a clean dictionary structure.
* Dynamically extracts available components (no hardcoded `Ez` assumptions).
* Precomputes the time axis and handles negative polarity suffixes directly.
* Includes a suite of 47 unit tests using an in-memory mocked HDF5 fixture, so it passes CI without needing real simulation files.

### 2. The Dashboard (`ascan_dashboard.py`)
This replaces the terminal workflow with a browser UI.
* **Trace Picker:** Users can load multiple files, pick specific receivers/components, and add them to a single plot.
* **Dual Y-Axis:** Solved the scaling conflict by plotting E-field (V/m) on the left (solid lines) and H-field (A/m) on the right (dashed lines). 
* **State Management:** Isolated the trace list state into its own cell so the UI doesn't drop traces when a dropdown value changes.
* **Exports:** Added buttons for CSV (full unzoomed data) and publication-quality SVG/PDF (requires `kaleido`).

## Screenshots
[GPRMax A-Scan.pdf](https://github.com/user-attachments/files/29965975/GPRMax.A-Scan.pdf)

Ascan Dashboard (6)
GPRMax A-Scan

## How to test locally
1. `pip install -r toolboxes/Marimo/requirements.txt`
2. Run `plotly_get_chrome` in your terminal (one-time setup for the SVG export).
3. `pytest tests/test_h5_reader.py -v` (Should see 47 passes).
4. `marimo run toolboxes/Marimo/ascan_dashboard.py` and load up `examples/cylinder_Ascan_2D.h5`.

---
*Part of GSoC 2026 Component 3.*

PR #795, opened 2026-08-20
Add processing module: gain, background removal, subtraction and FFT (Component 3) : GSoC 2026

alphaleporus:
Builds on #696 (A-scan dashboard + `h5_reader.py`) and #720 (B-scan dashboard). Both are unmerged, so their commits and files appear in this diff. New files here are `toolboxes/Marimo/processing.py`, `tests/test_processing.py` and `examples/cylinder_Ascan_2D_freespace.in`.

## Summary

`processing.py` is a marimo-free numpy module holding the amplitude processing for both dashboards. One implementation is used for a single A-scan trace and for a full B-scan matrix, so the two cannot drift apart.

**Gain.** Constant, linear, power (`t^b`), exponential (`exp(a·t)`), dB per ns, and SEC (`exp(a·t) · t^b`). SEC is the gain most published GPR flows use; `b=1` reduces its spreading term to the textbook linear gain. Gain onset is adjustable so the direct wave can be left alone instead of saturating the plot, and the curve can be bounded. The applied curve is plotted in its own panel below the figure in both dashboards.

**Free-space subtraction.** Run the model twice, once with the target and once without, and subtract. What cancels is everything the two runs share: the source pulse, the direct coupling between antennas, and any flat-interface reflection. What survives is the target. `examples/cylinder_Ascan_2D_freespace.in` is included so this is testable without constructing the reference run by hand. On the standard example the cylinder response is about 12% of the direct wave and effectively invisible; after subtraction it is the largest feature in the trace.

**Background removal.** Mean-trace subtraction across a B-scan, globally or over a moving window of traces. This is the equivalent of the above for real survey data, where no free-space run exists. Global is the default: every trace in a gprMax B-scan comes from the identical model, so the background is genuinely stationary. The window matches the convention for survey lines where it drifts along the profile.

**FFT.** Frequency-domain toggle on the A-scan dashboard, using `fft_power` from `gprMax/utilities/utilities.py` rather than reimplementing the transform.

Processing order is subtraction and background removal, then gain, then display normalisation. Gaining first amplifies the stationary direct wave and leaves a larger residual for the subtraction to cancel.

## Design notes

Subtraction is treated as a data operation and gain as a display one, which is why subtraction also applies to the spectrum and gain does not. Gain multiplies in time, which convolves in frequency, so a centre frequency read off a gained spectrum would be wrong. Subtraction isolates the signal you actually want to analyse, so its spectrum is the meaningful one.

CSV export always contains the untouched file contents. Every CSV writes a first line recording what was applied, including when nothing was, so the format never changes depending on a UI setting. In frequency domain the CSV holds the spectrum at full resolution regardless of the view range, and the first line says so.

`subtract_traces` checks the time step as well as the sample count. Two runs can produce the same iteration count from different `#time_window` or `#dx_dy_dz` settings, in which case the arrays line up index by index while representing different instants and the subtraction silently returns nonsense. Length alone would pass. A trace whose own file is the reference is left alone rather than zeroed, and a component missing from the reference leaves that trace unsubtracted with a note rather than failing the render.

AGC is deliberately absent. It equalises amplitudes over a local window and destroys relative reflectivity, which is the one thing an FDTD simulation gets right and a field instrument does not.

The power term is floored at 1.0 rather than evaluated as a bare `t^b`, which is zero at `t=0` and below 1 for all `t<1 ns`. On a 3 ns window that would blank the first sample and attenuate the first third of the trace. Flooring keeps `t^b` exact everywhere it amplifies.

Gain onset is stored as a percentage of the time window, not an absolute time. An ns-valued control would depend on the loaded data, so it would be rebuilt and reset whenever a trace was added or, on the live B-scan, whenever a new trace arrived mid-run.

## Notes on existing code

`fft_power` normalises each trace against its own maximum before returning, so every spectrum peaks at 0 dB regardless of absolute amplitude and two traces 20x apart plot identically when overlaid. The dashboard offers a shared reference that recovers the true ratio, computed within each unit family since a decibel difference between V/m and A/m is not a physical statement.

An all-zero trace produces `-inf` on the same path, which the function's own guard rewrites to 0 and the shift turns into a flat 0 dB line indistinguishable from real data. `Ex` in a 2D TMz model is exactly this case. The dashboard names such traces instead of drawing them.

The `freqmaxpower * 4` view fallback in `toolboxes/Plotting/plot_Ascan.py` is applied as an array index and can run past the end of the array when the peak sits beyond a quarter of the spectrum. `spectrum_view_limit()` does the same thing in frequency and clamps.

## Tests

68 tests in `tests/test_processing.py`. Guards are mutation-tested: removing the gain broadcast axis, the negative-gain clip, the gain-onset floor, the power floor, the moving-window edge clamp, the subtraction shape check or the subtraction time-step check each fails at least one named test.

`processing.py` has no marimo dependency and the `fft_power` import is deferred, so the module and most of its tests run on a checkout without a built gprMax. The FFT tests skip rather than fail.

## How to test

`plotly_get_chrome` is needed once for the SVG and PDF buttons. The camera icon in the plot toolbar works without it.

```
pytest tests/test_processing.py -v
```

**Gain and the curve panel**

```
python -m gprMax examples/cylinder_Ascan_2D.in
marimo run toolboxes/Marimo/ascan_dashboard.py
```

Load `cylinder_Ascan_2D.h5` and add `Ez`. Under Step 5 pick any gain. The curve appears below the plot on a log axis, the y-axis label gains a `(gained)` marker, and the title records the settings. At 6 dB per ns the curve reads 1.41 at 0.5 ns and 7.94 at 3 ns, which is `10^(6t/20)`. Raise the gain start and the curve stays flat at 1 before it. Download the CSV and confirm the first line records the gain while the values below are unchanged from the ungained export.

**Free-space subtraction**

```
python -m gprMax examples/cylinder_Ascan_2D_freespace.in
```

Load both `.h5` files, add `Ez` from `cylinder_Ascan_2D.h5`, then pick the free-space file in Step 1b. The direct wave near 1.1 ns cancels and the cylinder reflection near 2.2 ns becomes the largest feature. The title records the subtraction and the CSV first line says the exported values do not have it applied.

**FFT**

Add `Hx` and `Ex`, then switch Step 6 to Frequency. `Ex` is listed in a warning callout rather than drawn. The view auto-ranges to about 6.7 GHz out of a 105 GHz Nyquist and both spectra peak near 1.5 GHz, matching the source. Tick "Show full spectrum to Nyquist" to see why the auto-range exists.

**Background removal and gain on a radargram**

```
python -m gprMax examples/cylinder_Bscan_2D.in -n 60
marimo run toolboxes/Marimo/bscan_dashboard.py
```

Use Part 2 and select the 60 trace files. Under Step 10b tick background removal: the direct wave disappears and the hyperbola is left. Then set gain to SEC. The title records both operations in the order applied, and the CSV first line repeats it while the matrix below stays raw. Part 1 takes the same controls against a live run.

## Screenshots
Screenshot 2026-08-20 at 3 49 02 AM
Screenshot 2026-08-20 at 1 44 54 PM
Screenshot 2026-08-20 at 1 55 19 PM
Screenshot 2026-08-21 at 2 19 14 AM

### DIFF
--- a/examples/cylinder_Ascan_2D_freespace.in
+++ b/examples/cylinder_Ascan_2D_freespace.in
@@ -1,0 +1,13 @@
+#title: Free-space reference for cylinder_Ascan_2D, target removed
+#domain: 0.240 0.210 0.002
+#dx_dy_dz: 0.002 0.002 0.002
+#time_window: 3e-9
+
+#material: 6 0 1 0 half_space
+
+#waveform: ricker 1 1.5e9 my_ricker
+#hertzian_dipole: z 0.100 0.170 0 my_ricker
+#rx: 0.140 0.170 0
+
+#box: 0 0 0 0.240 0.170 0.002 half_space
+
--- a/tests/test_h5_reader.py
+++ b/tests/test_h5_reader.py
@@ -1,0 +1,342 @@
+"""
+tests/test_h5_reader.py: unit tests for toolboxes/Marimo/h5_reader.py
+
+Run from repo root:
+    pytest tests/test_h5_reader.py -v
+
+Uses a synthetic HDF5 fixture built in memory — no real gprMax output
+file required. Tests cover metadata extraction, component reading,
+utility functions, and error handling.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import h5py
+import numpy as np
+import pytest
+
+from toolboxes.Marimo.h5_reader import (
+    build_label,
+    format_metadata_text,
+    get_time_axis,
+    get_trace,
+    get_unit_label,
+    list_components,
+    list_receivers,
+    load_file,
+    load_files,
+)
+
+# Fixtures
+
+
+DT = 4.717308673499368e-12  # seconds (from real cylinder_Ascan_2D.h5)
+ITERATIONS = 637
+NRX = 1
+NSRC = 1
+
+COMPONENTS = {
+    "Ex": np.zeros(ITERATIONS, dtype=np.float64),
+    "Ey": np.zeros(ITERATIONS, dtype=np.float64),
+    "Ez": np.linspace(-1000, 1000, ITERATIONS, dtype=np.float64),
+    "Hx": np.ones(ITERATIONS, dtype=np.float64) * 0.5,
+    "Hy": np.ones(ITERATIONS, dtype=np.float64) * -0.5,
+    "Hz": np.zeros(ITERATIONS, dtype=np.float64),
+}
+
+
+def _write_synthetic_h5(path: Path, title: str = "Test A-scan") -> None:
+    """Write a minimal valid gprMax v4 HDF5 file for testing."""
+    with h5py.File(path, "w") as f:
+        # Root attributes
+        f.attrs["Title"] = title
+        f.attrs["dt"] = DT
+        f.attrs["dx_dy_dz"] = np.array([0.002, 0.002, 0.002])
+        f.attrs["Iterations"] = ITERATIONS
+        f.attrs["nrx"] = NRX
+        f.attrs["nsrc"] = NSRC
+        f.attrs["nx_ny_nz"] = np.array([120, 105, 1])
+        f.attrs["gprMax"] = "4.0.0b0"
+
+        # Receiver
+        rx1 = f.require_group("rxs/rx1")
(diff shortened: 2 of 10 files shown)


agianno:

PR #796, opened 2026-08-20
Add reactive recipes and hyperbola module (Component 4) : GSoC 2026

alphaleporus:
Builds on #696, #720 and #795. Those are unmerged, so their commits and files appear in this diff. New here are `toolboxes/Marimo/hyperbola.py`, `toolboxes/Marimo/recipes/`, and `tests/test_hyperbola.py`.

## Summary

Two self-contained teaching notebooks, plus the travel-time module they share.

**`recipes/ascan_workflow.py`** goes from parameters to waveform in one notebook: sliders drive a live `.in` preview, the file is written to disk, gprMax runs as a subprocess with live iteration progress, and the output is read back and plotted. Before the solver starts, the notebook predicts the direct wave and reflection arrivals from the geometry and warns if the reflection would fall outside the requested time window. Those predictions are then drawn on the measured trace.

**`recipes/velocity_permittivity.py`** overlays a predicted hyperbola on a real radargram. Moving the permittivity slider moves the curve; when it sits on the reflection, the slider reads the permittivity of the material. It also reads a permittivity back from the measured apex, which is the calculator direction.

**`hyperbola.py`** holds the maths: wave speed from permittivity and back, the gprMax ricker delay, two-way travel time for a bistatic pair against a target of finite radius, the apex position and time, and the inverse from a measured apex.

The recipes do not import the other notebooks. A marimo notebook is a `marimo.App` with `@app.cell` functions and no module-level API, so there is nothing importable in `parameter_controls.py` or `progress_tracker.py`. Both recipes rebuild the parts they need and reuse the pure modules (`h5_reader`, `trace_matrix`, `processing`, `hyperbola`) instead.

## Three things the travel-time model has to get right

**The source waveform delay.** gprMax evaluates the ricker at `t - sqrt(2)/f`, so a 1.5 GHz pulse leaves 0.943 ns after the simulation starts. That is 31% of the 3 ns window the standard examples use. A prediction without it lands a third of the record early. It is also not recoverable from the output file, which stores dt, iterations, grid and source position but not the source waveform, so it has to be supplied or estimated from the spectrum.

**The bistatic apex offset.** The hyperbola turns over when the midpoint of the transmitter and receiver crosses the target, half a separation before the source gets there. For `cylinder_Bscan_2D.in` that is 20 mm, and reading the apex straight off the source axis misplaces the target by that much.

**The target radius.** The reflection comes off the nearest point of the cylinder, not its centre. For the 10 mm cylinder in the standard example, ignoring this puts the arrival 0.163 ns late.

## Two limits the recipes state rather than hide

**Depth and permittivity cannot be solved together.** Fitting a 20-trace hyperbola with the source delay known recovers depth to half a millimetre and permittivity to 0.05. Letting the delay float as well, the same arrivals admit permittivities from roughly 4 to 7 at sub-picosecond residual. Depth is therefore an input to the velocity recipe, not an output.

**Background removal biases a picked arrival.** Subtracting the across-trace mean is what makes a hyperbola visible, but over a limited aperture the moving reflection does not average out, so the mean carries a smeared copy of it. On a 60-trace run of `cylinder_Bscan_2D.in` that pulls the picked apex 0.045 ns early and reports 5.6 against a modelled 6.0. Reading from the raw traces gives 6.02. The recipe processes for display and measures from raw data, and the readout does not change when the display toggles do.

Against that same run the model reproduces the measured arrivals to within about 4%, arriving slightly late. The gap is not tuned away: the model gives the geometric arrival of an idealised impulse, the data is the peak of a finite-bandwidth wavelet through a dispersive grid. Note that permittivity goes as the square of elapsed time, so a 2.6% timing error becomes 8.6% in the recovered value.

## Tests

26 tests in `tests/test_hyperbola.py`, with every constant read out of `examples/cylinder_Bscan_2D.in` rather than assumed. Includes a check against the measured arrivals from a real 60-trace run, and a test that fails if the source delay is dropped. Guards are mutation-tested: removing the radius correction, the bistatic offset, the `sqrt(2)/f` delay, the superluminal check or the depth check each fails at least one named test.

`hyperbola.py` has no marimo dependency and needs nothing but numpy.

## How to test

```
pytest tests/test_hyperbola.py -v
```

**End-to-end recipe**

```
marimo run toolboxes/Marimo/recipes/ascan_workflow.py
```

Press **Write and run** at the defaults. The generated input file matches 'examples/cylinder_Bscan_2D.in' on material, waveform, domain, geometry and antenna separation.. The predicted direct wave marker should land on the first event and the reflection marker on the second, each within a few percent.

Then set permittivity to 12. Step 2 warns before the run that the reflection will arrive at 3.24 ns and the window is only 3 ns. Run it anyway: the trace has no reflection and Step 5 says why.

**Velocity recipe**

```
python -m gprMax examples/cylinder_Bscan_2D.in -n 60
marimo run toolboxes/Marimo/recipes/velocity_permittivity.py
```

Select the 60 output files. The defaults already match that model, so the dashed curve should sit on the reflection with the slider at 6.0. Drag it to 3 and the curve lifts off the data; drag it back and it settles. Toggle background removal and SEC gain and confirm the readout does not move.

## Screenshots
Screenshot 2026-08-20 at 4 03 30 PM
Screenshot 2026-08-20 at 4 03 43 PM
Screenshot 2026-08-20 at 4 12 48 PM
Screenshot 2026-08-20 at 3 18 46 PM
Screenshot 2026-08-20 at 3 18 58 PM
Screenshot 2026-08-20 at 3 19 07 PM
Screenshot 2026-08-20 at 3 41 50 PM

### DIFF
--- a/examples/cylinder_Ascan_2D_freespace.in
+++ b/examples/cylinder_Ascan_2D_freespace.in
@@ -1,0 +1,13 @@
+#title: Free-space reference for cylinder_Ascan_2D, target removed
+#domain: 0.240 0.210 0.002
+#dx_dy_dz: 0.002 0.002 0.002
+#time_window: 3e-9
+
+#material: 6 0 1 0 half_space
+
+#waveform: ricker 1 1.5e9 my_ricker
+#hertzian_dipole: z 0.100 0.170 0 my_ricker
+#rx: 0.140 0.170 0
+
+#box: 0 0 0 0.240 0.170 0.002 half_space
+
--- a/tests/test_h5_reader.py
+++ b/tests/test_h5_reader.py
@@ -1,0 +1,342 @@
+"""
+tests/test_h5_reader.py: unit tests for toolboxes/Marimo/h5_reader.py
+
+Run from repo root:
+    pytest tests/test_h5_reader.py -v
+
+Uses a synthetic HDF5 fixture built in memory — no real gprMax output
+file required. Tests cover metadata extraction, component reading,
+utility functions, and error handling.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import h5py
+import numpy as np
+import pytest
+
+from toolboxes.Marimo.h5_reader import (
+    build_label,
+    format_metadata_text,
+    get_time_axis,
+    get_trace,
+    get_unit_label,
+    list_components,
+    list_receivers,
+    load_file,
+    load_files,
+)
+
+# Fixtures
+
+
+DT = 4.717308673499368e-12  # seconds (from real cylinder_Ascan_2D.h5)
+ITERATIONS = 637
+NRX = 1
+NSRC = 1
+
+COMPONENTS = {
+    "Ex": np.zeros(ITERATIONS, dtype=np.float64),
+    "Ey": np.zeros(ITERATIONS, dtype=np.float64),
+    "Ez": np.linspace(-1000, 1000, ITERATIONS, dtype=np.float64),
+    "Hx": np.ones(ITERATIONS, dtype=np.float64) * 0.5,
+    "Hy": np.ones(ITERATIONS, dtype=np.float64) * -0.5,
+    "Hz": np.zeros(ITERATIONS, dtype=np.float64),
+}
+
+
+def _write_synthetic_h5(path: Path, title: str = "Test A-scan") -> None:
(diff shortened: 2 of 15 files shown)


agianno:

PR #830, opened 2026-09-07
Fix release physics regressions in materials, sources and absorption

agianno:
## Summary

Follow-up release fixes from the numerical reviews after #829. This PR includes the implementation corrections, focused regressions, user documentation, and corrected grounded-dipole analytical references. It does not include temporary review reports or the separate Vivaldi/MATLAB comparison.

### Material identity and directional absorption

- Reserve `+` in user-defined material IDs to prevent collisions with generated interface/source materials. Apply the restriction to native/API/database declarations and STEP/STL material mappings; preserve internally generated names during geometry export/import.
- Retain the ordered x/y/z material definitions of diagonal-anisotropic volume cells. Compute electric absorption from the sum of component-specific losses, including dispersive poles, rather than a scalar mean.
- Preserve directional definitions through geometry-object read/write, material collection and MPI absorption payloads. SAR requires equal positive constituent densities; density and tag membership are not interface-averaged.
- Warn explicitly when SAR/radiometry selects magnetic materials: these outputs include electric absorption only. Magnetic properties still affect the field solver; magnetic heating is not implemented. Document the distinction between magnetic loss and lossless permeability.

### Sources and port outputs

- Rebind gathered MPI voltage-port outputs using their stable port IDs instead of position alone, so separately named coincident sources finalise correctly, including geometry reuse.
- Use the same full dispersive feed-gap admittance in native voltage-port S11/Zin and downstream antenna/SAR power calculations.
- Validate point sources against their actual staggered Yee-component bounds during initial placement, stepping and studies. Reject padded-array locations beyond the physical component while preserving valid terminal/symmetry-plane edges and the live 2-D layer.

### Spatial-resolution diagnostics

- Exclude the invariant-axis thickness from 2-D wavelength checks.
- Include electric and magnetic constitutive response in material wavelength/attenuation estimates. Distinguish these from the limited lossless-isotropic numerical phase-speed estimate and the existing conservative output-frequency guard.
- Exclude explicitly zero-amplitude waveforms from the pre-solve source-band diagnostic without removing passive source/port structures.
- Analyse waveforms independently; handle unavailable bandwidth estimates without indexing failures or suppressing a known under-resolution result. Combine diagnostic results and failures collectively under MPI.
- Add `--allow-underresolved` / `gprMax.run(..., allow_underresolved=True)` for intentional coarse-mesh experiments. The default remains rejection; the override does not bypass material stability, CFL or output-frequency validity checks.

### Analytical reference correction

- Position analytical grounded magnetic dipoles at their actual Yee centres, and regenerate the affected reference plots/metrics.
- Retain absolute complex-field comparisons, in addition to pattern comparisons. This corrects the reference model, not an identified field-solver error.

## Verification

Final local checks against commit `9a1447d7`:

| Check | Result |
| --- | --- |
| CI unit-test selection (Python 3.12 CPU/MPI environment) | **5,025 passed, 3 environment skips** |
| Focused real CPU/MPI and reference regressions | **158 passed** |
| Focused real CUDA/NVIDIA OpenCL regressions | **58 passed; no hardware skips** |
| Sphinx HTML build from a staged-only copy, with warnings treated as errors | **Passed without warnings** |
| Changed Python files parse; `git diff --check` | **Passed** |

The unit selection skips unavailable PyCUDA, pythonocc-core and PyVista in the CPU environment. CUDA/OpenCL execution was performed separately in the GPU environment (Python 3.13); the missing PyCUDA unit-environment dependency does not replace those hardware runs.

The focused runs cover physical source boundaries in single/double precision and multiple MPI decompositions; coincident voltage ports at rank interiors/faces in all orientations, automatic/explicit IDs and reused geometry; dispersive feed-power consistency; directional absorption and geometry import/export; all six reduced TE/TM orientations; passive-source diagnostic exclusion; under-resolution control; magnetic-absorption warnings; and corrected grounded-dipole reference checks.

Independent checks include Debye shunt admittance, homogeneous propagation/mesh refinement, and analytical dipole reference conventions. Backend and output-path agreement checks are consistency evidence, not a substitute for those independent comparisons. The warning-enabled/suppressed CPU, CUDA and OpenCL comparisons preserve the checked absorption arrays bit-for-bit.

The CPU/MPI regression selection excludes GPU tests, and the GPU selection uses actual CUDA/OpenCL devices. Metal execution remains a separate Mac-side check.

## Scope and remaining limitations

- No eigenmode/waveport implementation or accelerator field-update kernel is changed by this PR.
- Real CUDA and NVIDIA OpenCL hardware and local MPI ranks are exercised; Metal cannot be executed on this Linux server.
- Backend agreement and output-path consistency are distinguished from independent analytical/convergence checks in the tests and documentation.
- Magnetic absorption is warned about, not implemented. The under-resolution override does not certify an inaccurate model, and sampled dispersive diagnostics are not a general stability proof.


agianno:
## CI follow-up: 98c4fb86

The previous CPU job finished with 2 failures, 3475 passes and 49 skips. Both failures were in the new distributed directional-absorption test: it also requested collective geometry export using `driver="mpio"`, but the pip-installed h5py 3.16.0 in that job has no MPI support. This was reproduced locally. The other CI jobs on the previous commit passed.

Changes are confined to one test file and the CPU workflow:

- Split the test into absorption/reused-geometry and geometry export/import variants for both decompositions. Ordinary MPI SAR and radiometry comparisons still run with serial HDF5. Only export/import variants require MPI-enabled HDF5. No numerical assertions or tolerances were weakened.
- Use Open MPI's `ob1` transport with `self,sm,tcp` for this hosted CPU job. The failed job also logged UCX errors while probing Azure's network device; these job-only settings avoid requiring RDMA. Normal gprMax execution is unchanged.
- Stop forcing `FI_PROVIDER=shm` inside the new regression, matching the existing MPI release tests. A separate MPICH 4.1.1 diagnostic completed the simulation but hung in `MPI_Finalize` with this forced provider; the same model completed and finalized with default provider selection. Finalization remains enabled.

Local checks on the final test changes:

- Open MPI 5.0.8 / Python 3.12 / MPI-enabled HDF5: **27 passed, 2 deselected**, including all four absorption and geometry-roundtrip variants, plus magnetic-absorption warning and MPI test-environment checks.
- Open MPI / Python 3.12 / serial h5py **3.16.0**, matching the CI h5py version: **2 passed, 2 export-only skips**.
- MPICH 4.1.1 / Python 3.13 / serial HDF5: **2 passed, 2 export-only skips** after removing the forced provider.
- Python syntax, workflow YAML/settings and `git diff --check`: passed.

The full CPU CI selection (`not unit and not gpu and not slow`) completed locally with **3472 passed, 56 skipped, 5711 deselected, no failures** in 1542.61 seconds. It used the Python 3.12 environment with CI's serial h5py 3.16.0; the separate local Vivaldi test, which is not included in this PR, was excluded. Skips include optional dependencies, parallel-HDF5 tests, existing unimplemented placeholders, and five tests whose launcher lookup only checked alongside the virtual environment's Python executable.

After making the same Open MPI launcher available beside the temporary environment's Python executable, those five skipped propagation/dispersion-control cases were rerun: **5 passed in 11.13 seconds**. No repository change was needed for this temporary-environment layout.

New GitHub checks have been triggered. The unit-test workflow passed; CPU and macOS x86-64 wheel checks were still running when these local results were recorded. The Azure runner itself must confirm the hosted transport configuration.

No solver implementation, field update or numerical formula is changed in this follow-up.

PR #834, opened 2026-09-12
Added CUDA solver support for HSG sub-grids

Codebruh-sudo:
Sub-gridding was the last major feature restricted to the CPU solver. This adds a CUDA path, so a model containing SubGridHSG regions can run with -gpu.

All CUDA-specific code lives in new modules. The CPU classes are unchanged.

##What's new

Two kernel modules. knl_subgrid_hsg.py has the interface kernels (update_is, update_electric_os, update_magnetic_os), and knl_subgrid_precursors.py has gather_taps, bilinear_interp and time_blend.

Four host modules. cuda_subgrid_hsg.py adds CUDASubGridHSG, cuda_subgrid_interface.py drives the 24 IS/OS launches, cuda_precursor_nodes.py keeps the precursors on the device for both the filtered and unfiltered classes, and cuda_subgrid_updates.py provides the updaters.

##What changed in existing code

Four files, a few lines each.
config.py - the accelerator guard now blocks OpenCL and Metal only, since CUDA has sub-grid kernels and they don't.
solvers.py - the two isinstance(self.updates, SubgridUpdates) checks become hasattr checks. CUDASubgridUpdates subclasses CUDAUpdates, so isinstance was false on the GPU and the sub-grid phases were skipped silently. Also, set_dispersive_updates() is now CPU-only: the CUDA path picks its dispersive kernels when they're built, so calling it would replace them with Cython pointers.
subgrids/updates.py  -create_updates() dispatches to CUDA before reaching the CPU loop, which tests type(sg) == SubGridHSG and so rejects a subclass.
subgrids/user_objects.py- build() constructs a CUDASubGridHSG on the GPU. Imports are local so CPU-only installs never load PyCUDA.

##Implementation notes

interpolate_to_sub_grid uses RectBivariateSpline, which can't run on the device. A degree-1 spline on a rectilinear grid is bilinear and the coordinates are fixed at setup, so it reduces to four reads with precomputed weights. Checked against SciPy to 3e-16.
The 3-tap filter and the transverse blend are algebraically one four-tap weighted sum, so a single kernel serves both precursor classes with different constants.
Both grids stay on the device and share a CUDA context they exchange fields several times per main-grid step, so per-step host copies would remove the benefit.

CUDASubgridUpdater keeps its own copy of hsg_1 and hsg_2. It can't inherit them: those methods use zero-argument super(), which binds to SubgridUpdater's MRO and would send the bulk updates back to Cython. A small AST check compares the two and fails if they diverge; it has already caught the observe_sar_electric addition and the store_outputs reordering.

##Validation

Four configurations run on both backends, comparing every receiver dataset. Figures are the largest difference divided by that dataset's peak. T4, double precision.
Debye water cylinder: 2.7e-10 main grid, 2.7e-13 sub-grid. 2 m 03 s CPU, 11.0 s GPU.
Two sub-grids in one model: 2.0e-10 main grid, 1.7e-13 and 4.0e-13 sub-grids. 3 m 48 s, 19.8 s.
filter=False: 2.7e-10 and 1.9e-13. 2 m 03 s, 11.7 s.
Ratio 3: 2.8e-10 and 1.8e-13. 33.0 s, 16.9 s.
Every model has a six-component receiver in the main grid as well as inside each sub-grid, so the outer-surface corrections are checked in both directions.
The main-grid figure is always Hz, whose peak is ~2e5 against 1e8–1e12 for the other components. It's the near-null component on the dipole axis, so it's mostly cancellation in absolute terms it agrees better than anything else.
Speed-up scales with the work inside the sub-grid. Ratio 3 gains least because the interface kernels launch just as often for far fewer cells.

##Not covered

PrecursorNodesEqualResolution (ratio 1) has no GPU equivalent yet, so the CUDA path raises NotImplementedError rather than running the wrong precursors. There's no interpolation involved, so it should be easy to add.
MPI with sub-grids is unsupported upstream on any backend, and OpenCL and Metal remain blocked.


agianno:
Added the reviewed fixes and regression coverage in two follow-up commits:

- `92aee1c6` — CUDA HSG parity, interpolation, virtual guides and solver lifecycle.
- `ef5a466e` — native transmission-line history/plotting compatibility.

### Changes

- Construct virtual-waveguide auxiliary grids/updaters using the ordinary backend, sharing the owning CUDA context and retaining the local timestep.
- Refresh context-bound material tables on geometry reuse; clean up failed main/fine construction safely.
- Implement the CPU's selectable spline degrees 1–5 on CUDA. Degrees 2–5 use precomputed separable FITPACK interpolation matrices; fields stay on the GPU. Retain filtered linear interpolation as the default, and document higher orders as experimental.
- Enable equal-resolution (`ratio=1`) CUDA regions in single/double precision, with direct co-located field transfer.
- Remove volume-sized temporary index arrays during precursor setup; use portable 64-bit descriptor arguments and pointer-sized flattened indexing.
- Read only physical samples from native TL allocation buffers, while preserving merged matrices and rejecting malformed histories. No source timing or HDF5 writer schema changes.
- Update capability documentation and add 149 regression cases plus a dielectric-sphere validation driver.

### Validation

Fresh pre-publication checks: **131 CPU-side checks passed and 20 real CUDA smoke cases passed**. The latter cover all interpolation orders, randomized material-index/interface checks, geometry reuse, both snapshot storage modes and single/double ratio-one virtual guides.

The larger validation campaign used four local TITAN RTX GPUs:

- 40 feature cases / 2,759 output datasets: no structural, non-finite-mask or checked timing/identity metadata mismatches. Worst receiver-family normalized differences: E `3.56e-13`, H `2.59e-12`, current `1.11e-11`.
- Ratios 1/3/5/7; filtered/unfiltered; multiple subgrids; sources/ports, virtual guides and modal studies; dispersive and magnetic media; internal PML; hard-source startup; snapshots; SAR, NTFF and output processing.
- CPU/CUDA analytic dipole error: approximately `-39.421868 dB` on both.
- CUDA memory checking of randomized precursor/interface kernels: zero errors.
- Ordinary OpenCL virtual-guide comparison also passes; this does not add OpenCL HSG support.

All six dielectric-sphere runs pass the unchanged Mie limits (RMS <=0.75 dB, maximum <=1.25 dB). Same physical sphere: radius 16 mm, relative permittivity 4, 0.75–9 GHz:

| Discretisation | CPU/CUDA RMS error vs Mie | CPU/CUDA maximum error |
|---|---:|---:|
| Uniform 0.5 mm | 0.270804 dB | 0.724875 dB |
| HSG linear, 1.5/0.5 mm | 0.398413 dB | 0.763582 dB |
| HSG cubic, 1.5/0.5 mm | 0.398886 dB | 0.770888 dB |

CPU/CUDA RCS differences are near round-off. Cubic offers no meaningful accuracy gain in this example.

Isolated benchmark, median of three CPU/CUDA pairs: 220-cubed main / 258-cubed fine grid including extensions, ratio 3, 80 coarse steps, double precision. CPU uses 20 OpenMP threads, CUDA one TITAN RTX. Solve time is **22.410 s CPU / 2.144 s CUDA (10.45x)**; full API time **26.141 s / 11.455 s (2.28x)**. These are workload-specific, warm-cache timings, not full-card capacity measurements.

### CI and local environment qualifications

The standard CPU selection was completed through split runs and transport-controlled MPI reruns: **4,582 passed / 50 skipped**, with a fresh collection audit finding no missing or unresolved selected cases. This was not one uninterrupted default-environment green run: local MPICH 4.1.1 hangs in `MPI_Finalize()` with `FI_PROVIDER=shm`, reproducible using mpi4py alone. TCP resolves it; all 26 SAR/voltage MPI assertions pass with a temporary TCP launcher, without weakening assertions or changing repository tests. The GitHub CPU workflow uses Open MPI.

The local `unit and not gpu` run gave **5,241 passed / 4 failed / 7 skipped**. The four failures are the previously identified local mpi4py-fft/FFTW 3-D plan-creation problem. They were not suppressed. Documentation builds successfully apart from one pre-existing title-underline warning.

Fresh GitHub CI is required for these commits before merging. Windows CUDA, Metal hardware and the remote A100 were not exercised locally. The supported HSG execution scope remains CPU/CUDA, not OpenCL, Metal or distributed MPI; these tests do not prove arbitrary long-time stability for experimental higher-order interpolation.

Codebruh-sudo:
Thanks for going through it and applying the fixes, they all make sense.  Congratulations on the 4.0.0 relea.e.

